PR #41, opened 2020-01-03
Fix syntax in example page that is incompatible with IE11

allen-garvey:
Removes ES2017 syntax for trailing commas in function calls from the example page, as this breaks on IE11.


davidcetinkaya:
Hi Allen (@allen-garvey),
Nice catch. And thank you very much for taking time to fix this. Much appreciated 🙂!

Best,
David

PR #152, opened 2021-03-04
Fix IE bug

sedlukha:
IE11 doesn't support node.contains (https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)
Breaks with error "Unable to get property 'contains' of undefined or null reference"

### DIFF
--- a/src/vanilla/components/utils.ts
+++ b/src/vanilla/components/utils.ts
@@ -52,11 +52,13 @@ export function arrayKeys<GenericType>(array: GenericType): number[] {
 }
 
 export function removeClass(node: HTMLElement, className: string): void {
-  const cl = node.classList
-  if (className && cl.contains(className)) cl.remove(className)
+  const cl = node.getAttribute('class')
+  if (className && cl && cl.indexOf(className) > -1)
+    node.setAttribute('class', cl.replace(className, ''))
 }
 
 export function addClass(node: HTMLElement, className: string): void {
-  const cl = node.classList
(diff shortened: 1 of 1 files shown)


davidcetinkaya:
Hi Arthur (@sedlukha),

First off, thank you very much for taking the time to make a contribution 👍. I couldn't reproduce the error in IE11 using BrowserStack. How were you able to provoke the error you mention?

> "Unable to get property 'contains' of undefined or null reference"

Looking at the support table for [classList](https://caniuse.com/classlist), IE11 supports both the `classList.add` and `classList.remove` methods:

classList support

Best,
David

sedlukha:
@davidcetinkaya I found this error with chromatic (https://www.chromatic.com/) cross-browser testing.
I don't know why it happens, but in browserstack everything works correct for me too
image

davidcetinkaya:
Thank you for providing the full error message @sedlukha. Seems like it tries to access `.contains` of something that's either `null` or `undefined` 😕. Do you know which **page/url** that triggers this error? 

Just guessing here but it possible that **Chromatic** tries to run the script before the carousel HTML is there?

sedlukha:
@davidcetinkaya 
Oh yes, you are right. Embla tried to add **selectedClass** before slider html was defined. Thanks for help. My PR lost any sense.

davidcetinkaya:
@sedlukha thank you anyway 👍. Feel free to make a contribution anytime if you find anything that needs to be fixed.

Cheers!
David

PR #222, opened 2021-09-23
Allow to assign slide nodes as slots within a web component using shadow DOM

klipstein:
Thanks for this great carousel library.

I would like to build my own `<my-embla-carousel>` components, which then could be used as such:

```html
<my-embla-carousel index="1">
  <my-slide slot="slide">Slide 1</my-slide>
  <my-slide slot="slide">Slide 2</my-slide>
</my-embla-carousel>
```

This pull request slightly adjusts how the slide children are collected when a slot element is involved.

### DIFF
--- a/packages/embla-carousel/package.json
+++ b/packages/embla-carousel/package.json
@@ -69,5 +69,8 @@
     "shx": "^0.3.3",
     "ts-jest": "^26.5.5",
     "typescript": "^4.2.4"
+  },
+  "dependencies": {
+    "@webcomponents/webcomponentsjs": "^2.6.0"
   }
 }
--- a/packages/embla-carousel/src/__tests__/embla-carousel-vanilla/index.test.ts
+++ b/packages/embla-carousel/src/__tests__/embla-carousel-vanilla/index.test.ts
@@ -3,6 +3,11 @@ import EmblaCarousel, {
   EmblaCarouselType,
 } from '../../embla-carousel-vanilla/index'
 
+// required when transpiling to es5 and web-components are used
+// otherwise the following error is thrown:
+// Class constructor HTMLElement cannot be invoked without 'new'
+import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js'
+
 const initializeEmbla = (
   withContainer: boolean = true,
   withSlides: boolean = true,
@@ -19,6 +24,23 @@ const initializeEmbla = (
   }
   return () => EmblaCarousel(rootNode, options)
 }
+
+class EmblaCarouselComponent extends HTMLElement {
+  carousel?: EmblaCarouselType
+
+  connectedCallback() {
+    const shadowRoot = this.attachShadow({ mode: 'open' })
+    const rootNode = document.createElement('div')
+    const containerNode = document.createElement('slot')
+    containerNode.setAttribute('name', 'slide')
+
+    rootNode.appendChild(containerNode)
+    shadowRoot.appendChild(rootNode)
+
+    this.carousel = EmblaCarousel(rootNode)
+  }
+}
+window.customElements.define('embla-carousel-component', EmblaCarouselComponent)
 
 describe('EmblaCarousel', () => {
   describe('When initialized it does not throw when', () => {
@@ -42,6 +64,24 @@ describe('EmblaCarousel', () => {
 
     test('The container node is not provided', () => {
       expect(initializeEmbla(false, false)).toThrow()
+    })
+  })
+
+  describe('When initialized inside web component with slot', () => {
+    test('It uses the assigned slots as slide nodes', () => {
+      const carouselNode = document.createElement(
+        'embla-carousel-component',
+      ) as EmblaCarouselComponent
+      for (let i = 0; i < 3; i += 1) {
+        const slide = document.createElement('div')
+        slide.setAttribute('slot', 'slide')
+        carouselNode.appendChild(slide)
+      }
+      document.body.appendChild(carouselNode)
+      // wait until componenet got attached
+      return Promise.resolve().then(() => {
+        expect(carouselNode.carousel?.slideNodes().length).toBe(3)
+      })
     })
   })
 })
--- a/packages/embla-carousel/src/embla-carousel-vanilla/index.ts
+++ b/packages/embla-carousel/src/embla-carousel-vanilla/index.ts
@@ -51,11 +51,18 @@ function EmblaCarousel(
 
   function setupElements(): void {
     if (!sliderRoot) throw new Error('Missing root node 😢')
-    const sliderContainer = sliderRoot.querySelector('*')
+    const sliderContainer = sliderRoot.querySelector<
+      HTMLElement | HTMLSlotElement
+    >('*')
     if (!sliderContainer) throw new Error('Missing container node 😢')
 
     container = sliderContainer as HTMLElement
-    slides = Array.prototype.slice.call(container.children)
+    if (sliderContainer instanceof HTMLSlotElement) {
+      slides = Array.prototype.slice.call(sliderContainer.assignedElements())
+    } else {
+      slides = Array.prototype.slice.call(container.children)
+    }
+
     optionsPseudo = OptionsPseudo(sliderRoot)
   }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -2368,6 +2368,11 @@
     "@webassemblyjs/ast" "1.11.0"
     "@xtuc/long" "4.2.2"
 
+"@webcomponents/webcomponentsjs@^2.6.0":
+  version "2.6.0"
+  resolved "https://registry.yarnpkg.com/@webcomponents/webcomponentsjs/-/webcomponentsjs-2.6.0.tgz#7d1674c40bddf0c6dd974c44ffd34512fe7274ff"
+  integrity sha512-Moog+Smx3ORTbWwuPqoclr+uvfLnciVd6wdCaVscHPrxbmQ/IJKm3wbB7hpzJtXWjAq2l/6QMlO85aZiOdtv5Q==
+
 "@xtuc/ieee754@^1.2.0":
   version "1.2.0"
   resolved "https://registry.yarnpkg.com/@xtuc/ieee754/-/ieee754-1.2.0.tgz#eef014a3145ae477a1cbc00cd1e552336dceb790"


davidcetinkaya:
Hi @klipstein,

Thank you for your contribution. So just a friendly reminder here, this is what the [contribution guidelines](https://github.com/davidcetinkaya/embla-carousel/blob/master/.github/CONTRIBUTING.md#new-features) say about **new features**:

- `Discuss:` Open an issue before starting any significant work. Let's discuss to see how the proposed feature fits within Embla Carousels's vision.

If you look at this [Pull Request](https://github.com/davidcetinkaya/embla-carousel/pull/220), it will finalise the monorepo setup separating the `embla-carousel` from the `embla-carousel-react` package. I'm thinking that we should aim to create a separate package for web components and call it `embla-carousel-webcomponents` or similar.

What's your thoughts about this?

Best,
David

klipstein:
Sorry for not reading your contribution guidelines before. My intention is not to propose a concrete embla web component and rather just allow embla to be used inside a webcomponent. As web-components are just an extension to the web platform I don't see it much different to what vanilla `embla-carousel` already provides.

In the future I think there could be the possibility to write an `embla-carousel-webcomponent` which offers a full fledged component and then it would make sense to create a separate package. But I guess this is way more work (e.g. allow passing options as attributes, emitting DOM events, documentation, ...) and depends on where you want to go with this library.

Here just a simplistic form of how I would write a custom web component using [lit](https://lit.dev/). Maybe just providing an example in the docs would be sufficient?

~~~html
<script type="module">
import EmblaCarousel from 'https://unpkg.com/embla-carousel/embla-carousel.umd.js';

class MyEmblaCarousel extends LitElement {
  static get properties() {
    return {
      playlistIndex: {
        type: Number,
        attribute: 'playlist-index'
      }
    };
  }

  firstUpdated() {
    const emblaNode = this.shadowRoot.querySelector('.embla__viewport');
    const emblaContainer = this.shadowRoot.querySelector('.embla__container');
    const options = { loop: false };
    this.carousel = EmblaCarousel(emblaNode, options);
    this.carousel.scrollTo(this.playlistIndex);

    emblaContainer.addEventListener('slotchange', () => this.carousel.reInit());
  }

  disconnectedCallback() {
    if (this.carousel) {
      this.carousel.destroy();
    }
  }

  updated(changedProperties) {
    if (this.carousel && changedProperties.has('playlistIndex')) {
      this.carousel.scrollTo(this.playlistIndex);
    }
  }

  _click() {
    if (!this.carousel.clickAllowed()) return;
    this.dispatchEvent(
      new CustomEvent('carouselclick', {
        bubbles: true,
        composed: true
      })
    );
  }

  render() {
    return html`
      <style>
        .embla__viewport {
          overflow: hidden;
        }
        .embla__container {
          display: flex;
        }
        .embla__container::slotted(*) {
          position: relative;
          flex: 0 0 100%;
        }
      </style>
      <div class="embla">
        <div class="embla__viewport">
          <slot name="slide" class="embla__container" @click=${this._click}></slot>
        </div>
      </div>
    `;
  }
}
window.customElements.define('my-embla-carousel', MyEmblaCarousel);
</script>
<my-embla-carousel>
  <div slot="slide>Slide 1</div>
  <div slot="slide>Slide 2</div>
  <div slot="slide>Slide 3</div>
</my-embla-carousel>
```

**Comment by davidcetinkaya** [comment]
@klipstein thank you for the additional information and the example code!

> In the future I think there could be the possibility to write an embla-carousel-webcomponent which offers a full fledged component and then it would make sense to create a separate package. But I guess this is way more work (e.g. allow passing options as attributes, emitting DOM events, documentation, ...) and depends on where you want to go with this library.

Please forgive me if this is a stupid question (I have zero web components and lit experience).

Is it possible to create a **minimal wrapper** like the `embla-carousel-react` version that **gives users maximum freedom** but still **makes the process of using Embla Carousel smooth**?

Here's how Embla Carousel is used with React:

```jsx
import React, { useEffect } from 'react'
import { useEmblaCarousel } from 'embla-carousel/react'

export const EmblaCarousel = (props) => {
  const { slides } = props
  const [emblaRef, emblaApi] = useEmblaCarousel({ loop: false }) // Embla options goes here

  useEffect(() => {
    if (emblaApi) {
      // Make use of Embla API here
    }
  }, [emblaApi])

  return (
    <div className="embla" ref={emblaRef}> {/* Attach a reference to the HTML element for React to use */}
      <div className="embla__container">
        {slides.map(slide => <div className="embla__slide">{slide}</div>}
      </div>
    </div>
  )
}
```

Cheers,
David

klipstein:
In my opinion creating a minimal `embla-carousel-webcomponent` would not suffice how you would use a web component and how you use a web component to hide details. Of course you can simply wrap embla in a web-component and just expose the `emblaApi`, but this is not how web-components are commonly used.

I can try to explain the differences a little bit how things would be different in regards to the API when you would have a web component (assuming you named your web component `embla-carousel-webcomponent`):

- instead of doing `emblaCarousel.on(...)` you would use the built-in EventTarget.addEventListener API: `document.querySelector('embla-carousel-webcomponent').addEventListener(...)`
- instead of passing in options via `new EmblaCarousel(node, { loop: true, dragFree: true})` you would pass those properties via DOM attributes: `<embla-carousel-webcomponent loop dragFree>` and react on changes of those within your component (see [attributeChangedCallback](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements))
- you could expose additional methods on your web-component if you want (similar to e.g. the `<video>` tag with `document.querySelector('video').play()` you could expose `document.querySelector('embla-carousel-webcomponent').scrollNext()` but it is more preferred that you e.g. allow to set the current slide index via an additional DOM attribute so that you can simply do a `document.querySelector('embla-carousel-webcomponent').setAttribute('index', 1)` and the component would then react on that.

Already today it is possible to create a web component without [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) and implement it as described above.

---

But without the changes I propose around considering shadow DOM slots as slides it is currently not possible to hide styling details within a component. With shadow DOM you can keep the style definitions like `.embla { margin-right: 20px; }` local to your component and this style definition would not affect things outside your component. That the `embla__` classes clash with other things on a page is probably an edge case but it would be good if this advanced way of styling would be possible.

The change would open additional things like communicating between LightDOM and ShadowDOM via `<slot>` and exposing the parts you want to expose via [part](https://www.thinktecture.com/de/web-components/native-web-components-without-framework/#css-shadow-parts) or that you granularly define which CSS properties can be overwritten via CSS variables and which not:

~~~css
embla-carousel-webcomponent {
  --background: red;
}
~~~

----

As far as I understand the concept of embla, the author has to do the slide / carousel styling on his own and I really like that it is this low level. In my case I would like to offer some web components, where the styling is kind of limited and a little bit opinionated so that you don't have to code too much to get a specific carousel on your page.

Because of that it would be great, if you could consider allowing an HTMLSlotElement as slider container. This would allow others to create customized carousel web components, that use shadow DOM. I'm not sure whether providing a too specific web component would help embla because you want to give an author the full power of styling.

Hope I was able to shed some light, why this feature would be useful without introducing a complete embla web component.

davidcetinkaya:
Hi @klipstein,

Thanks a lot for explaining this [in detail](https://github.com/davidcetinkaya/embla-carousel/pull/222#issuecomment-925877874). I've looked into the code changes you made in this PR:

```js
connectedCallback() {
  const shadowRoot = this.attachShadow({ mode: 'open' })
  const rootNode = document.createElement('div')
  const containerNode = document.createElement('slot')
  containerNode.setAttribute('name', 'slide')

  rootNode.appendChild(containerNode)
  shadowRoot.appendChild(rootNode)

  this.carousel = EmblaCarousel(rootNode)
}
```

And one question has come up. What if we made it possible to do this:

```js
const rootNode = document.createElement('div') // my custom root node
const containerNode = document.createElement('div')  // my custom container node
const slideNodes = [document.createElement('div'), document.createElement('div')] // my custom slide nodes

const options = { loop: false }
const embla = EmblaCarousel({ rootNode, containerNode, slideNodes }, options)
```

And still support the original way of doing it:

```js
const options = { loop: false }
const embla = EmblaCarousel(rootNode, options)
```

Wouldn't that enable us to create a small wrapper package that covers this use case?

Best,
David

davidcetinkaya:
Hi @klipstein,

Thanks a lot for explaining this [in detail](https://github.com/davidcetinkaya/embla-carousel/pull/222#issuecomment-925877874). Following up on your explanation:

> Of course you can simply wrap embla in a web-component and just expose the emblaApi, but this is not how web-components are commonly used.

Thanks to your explanation, I understand that this isn't how web components are commonly used. But the Embla Carousel vision is the following:

> It aims to solve the hardest technical challenges with building carousels, and the rest is up to the user utilizing its highly extensible API.

So creating a big component kind of goes against the Embla vision 😕. I've looked into the code changes you made in this PR:

```js
connectedCallback() {
  const shadowRoot = this.attachShadow({ mode: 'open' })
  const rootNode = document.createElement('div')
  const containerNode = document.createElement('slot')
  containerNode.setAttribute('name', 'slide')

  rootNode.appendChild(containerNode)
  shadowRoot.appendChild(rootNode)

  this.carousel = EmblaCarousel(rootNode)
}
```

And one question has come up. What if we made it possible to do this:

```js
const rootNode = document.createElement('div') // my custom root node
const containerNode = document.createElement('div')  // my custom container node
const slideNodes = [document.createElement('div'), document.createElement('div')] // my custom slide nodes

const options = { loop: false }
const embla = EmblaCarousel({ rootNode, containerNode, slideNodes }, options)
```

And still support the original way of doing it:

```js
const options = { loop: false }
const embla = EmblaCarousel(rootNode, options)
```

Wouldn't that enable us to create a small wrapper package that covers this use case? The only difference is that the few lines of code is something that the wrapper package would care of, instead of building this highly specific use case into the Embla Carousel core?

Best,
David

klipstein:
I was offline last week.

Allowing this in the API (what you explained) would also work for this usecase: `const embla = EmblaCarousel({ rootNode, containerNode, slideNodes }, options)`.

This way the magic of how the slides are picked up will be on the integration side then.

But having an embla web component within this repository by just exposing `component.carousel` doesn't make sense to me. Then you should rather just stick to vanilla DOM, which embla already provides.

Do you want to extend embla with this new "constructor" or shall I open a pull for it? I assume this pull can be closed then?

klipstein:
@davidcetinkaya there is no immediate need for the constructor change. For now I can create web components based on the fork I've created.

The interesting part of a web component is that it would automatically be compatible with React, Vue or Svelte when all the details would be configurable via attributes and it would expose standard events. This tweet also summarizes it quite well: https://twitter.com/justinfagnani/status/1444739792017653763?s=20

Because of that it would be great to have full-fledged embla web component at some point. I can provide some examples when I'm further into developing carousel components with embla.

I also checked whether there are other carousel JS projects that faced the same issue with slot projection and it seems that glide had a similar issue and someone solved it by creating individual slots per slide entry: https://github.com/glidejs/glide/issues/310

Closing this ticket now.

davidcetinkaya:
Hello @klipstein,

Just wanted to let you know that I just released Embla v6, which allows you to provide custom elements.

- [Here’s the commit](https://github.com/davidcetinkaya/embla-carousel/pull/229/commits/2498e6fcaa9e86ba1b6f94aa03987497cd5bcef6).
- [Release details here](https://github.com/davidcetinkaya/embla-carousel/releases/tag/v6.0.0)

Cheers,
David

klipstein:
Great. Will give it a try next week.

davidcetinkaya:
@klipstein have you had the chance to try [this](https://github.com/davidcetinkaya/embla-carousel/pull/222#issuecomment-962014906) yet?

Best,
David

klipstein:
Hi @davidcetinkaya,

I already tried it and it works great. Here is an extract of what I've implemented. The full example has previous/next buttons, a story-like progress indicator and it allows assigning slides which themself provide play / pause. This is where I've realized that it is not easy to write a generic carousel:

```js
import Embla from 'https://unpkg.com/embla-carousel@6.0.0/embla-carousel.esm.js';
import {
  html, LitElement
} from 'https://cdn.jsdelivr.net/npm/lit/+esm';

let idCounter = 0;
class MyCarousel extends LitElement {
  constructor() {
    super();
    this._onClick = this._onClick.bind(this);
    this.slidesToScroll = 1;
    this._carouselId = `slides-container-${idCounter}`;
    this.index = 0;
    idCounter += 1;
  }

  static get properties() {
    return {
      index: {
        type: Number
      },
      carouselIndex: {
        type: Number,
        reflect: false
      },
      canScrollPrev: {
        type: Boolean,
        reflect: false
      },
      canScrollNext: {
        type: Boolean,
        reflect: false
      },
      axis: {
        type: String
      },
      align: {
        type: String,
        converter: (value) => {
          const newValue = Number(value);
          return Number.isNaN(newValue)
            ? value
            : newValue;
        }
      },
      direction: {
        type: String
      },
      slidesToScroll: {
        type: Number,
        attribute: 'slides-to-scroll'
      },
      nonDraggable: {
        type: Boolean,
        attribute: 'non-draggable'
      },
      dragFree: {
        type: Boolean,
        attribute: 'drag-free'
      },
      loop: {
        type: Boolean
      },
      speed: {
        type: Number
      },
      containScroll: {
        type: String,
        attribute: 'contain-scroll'
      },
      skipSnaps: {
        type: Boolean,
        attribute: 'skip-snaps'
      },
      inViewThreshold: {
        type: Number,
        attribute: 'in-view-threshold'
      }
    };
  }

  get options() {
    let currentIndex = Math.floor(this.index / this.slidesToScroll);
    if (this.carousel) {
      currentIndex = this.carousel.selectedScrollSnap();
    }
    // remove undefined properties
    return JSON.parse(JSON.stringify({
      axis: this.axis,
      align: this.align,
      direction: this.direction,
      slidesToScroll: this.slidesToScroll,
      draggable: !this.nonDraggable,
      dragFree: this.dragFree,
      loop: this.loop,
      speed: this.speed,
      containScroll: this.containScroll,
      skipSnaps: this.skipSnaps,
      inViewThreshold: this.inViewThreshold,
      startIndex: currentIndex
    }));
  }

  _onClick(event) {
    if (!this.carousel.clickAllowed()) {
      event.stopPropagation();
    }
    this.focus();
  }

  focus() {
    if (!this.carousel) return;
    this.renderRoot.querySelector('.viewport').focus();
  }

  _initialSetup() {
    const root = this.renderRoot.querySelector('.viewport');
    const container = this.renderRoot.querySelector('.slides');
    if (this.carousel) {
      this.carousel.destroy();
    }
    this.carousel = Embla({
      root,
      container,
      slides: container.assignedElements()
    }, this.options);

    this._connectEmblaEvents();

    // stop bubbling click event while scrolling
    this.addEventListener('click', this._onClick, { capture: true });
  }

  firstUpdated() {
    const rootNode = this.renderRoot.querySelector('.viewport');
    if (!rootNode) return;

    const slidesNodes = this.renderRoot.querySelector('.slides');
    // when slides get assigned later
    slidesNodes.addEventListener('slotchange', () => {
      this._initialSetup();
    });
    if (slidesNodes.assignedElements().length > 0) {
      this._initialSetup();
    }
  }

  updated(changedProperties) {
    if (changedProperties.has('carouselIndex')
      || changedProperties.has('canScrollPrev')
      || changedProperties.has('canScrollNext')
    ) return;

    this.reInit();
    if (this.carousel && changedProperties.has('index')) {
      this.carousel.scrollTo(Math.floor(this.index / this.slidesToScroll));
    }
  }

  connectedCallback() {
    super.connectedCallback();
    this.firstUpdated();
  }

  disconnectedCallback() {
    if (this.carousel) {
      this.carousel.destroy();
      this.carousel = undefined;
    }
  }

  reInit() {
    if (!this.carousel) return;
    this.carousel.reInit(this.options);
  }

  _select() {
    this.canScrollPrev = this.carousel.canScrollPrev();
    this.canScrollNext = this.carousel.canScrollNext();
    this.carouselIndex = this.carousel.selectedScrollSnap() * this.slidesToScroll;
  }

  _init() {
    this._select({ initialSelect: true });
  }

  _connectEmblaEvents() {
    this.carousel.on('select', () => this._select());
    this.carousel.on('init', () => this._init());
  }

  render() {
    const slideChildren = this.querySelectorAll('[slot="slide"]');
    return html`
      <style>
        :host {
          position: relative;
          max-width: 100%;
          margin-left: auto;
          margin-right: auto;
          display: block;
        }

        .viewport {
          overflow: hidden;
          width: 100%;
          cursor: move;
          cursor: grab;
        }

        .viewport.hidden {
          visibility: hidden;
        }

        .slides {
          display: flex;
          user-select: none;
          -webkit-touch-callout: none;
          -khtml-user-select: none;
          -webkit-tap-highlight-color: transparent;
        }

        :host([non-draggable]) .viewport {
          cursor: initial;
        }
      </style>
      <div
        class="viewport${slideChildren.length === 0 ? ' hidden' : ''}"
      >
        <slot name="slide" class="slides" id="${this._carouselId}"></slot>
      </div>
    `;
  }
}
window.customElements.define('my-carousel', MyCarousel);
```

The `my-caoursel` can then be used like this:

```html
<style>
  my-carousel img[slot="slide"] {
        position: relative;
        flex-grow: 0;
        flex-shrink: 0;
        flex-basis: 100%;
        margin-right: 10px;
        overflow: hidden;
        max-width: 100%;
        background: rgb(180, 180, 180);
      }
</style>
<my-carousel loop>
  <img slot="slide" src="my-image1.png"/>
  <img slot="slide" src="my-image2.png"/>
  <img slot="slide" src="my-image3.png"/>
  <img slot="slide" src="my-image4.png"/>
</my-carousel>
```

davidjerleke:
Hi @klipstein,

This feature has now been released with [v7.1.0](https://github.com/davidjerleke/embla-carousel/releases/tag/v7.1.0) and is part of the official [options](https://www.embla-carousel.com/api/options/) API.

Best,
David

Issue #239, opened 2021-11-09
Bring the selected class back?


NachoVazquez:
### Feature request is related to

- [ ] The Vanilla JavaScript version
- [x] The React version
- [ ] The Embla Carousel core (all versions)
- [ ] Documentation website
- [ ] Other

### Is your feature request related to an issue?

I tried to update to the latest version but now my project is working as expected because of the removal of the selected class.

Is there any heavy reason why this was removed? Is there any chance to bring it back?

Is there any alternative to accomplish the same result?

### Describe the solution you'd like

- A restoration of the selected class functionality or an alternative

reco:
hey @NachoVazquez i think this is becoming a plug in https://github.com/davidcetinkaya/embla-carousel/issues/237

NachoVazquez:
> hey @NachoVazquez i think this is becoming a plug in https://github.com/davidcetinkaya/embla-carousel/issues/237

Thanks @reco ! That makes sense then!

davidcetinkaya:
Hi @NachoVazquez and @reco,

@reco, thanks for your [helpful response](https://github.com/davidcetinkaya/embla-carousel/issues/239#issuecomment-965731314)!

As @reco mentioned, I'm thinking about introducing the selected class logic in a plugin. Out of curiosity, do you think `draggableClass` and `draggingClass` is useful too? @reco feel free to leave your opinion here too.

Best,
David

RectoVersoDev:
Hi,

For me the classNames `.is-draggable` & `.is-dragging` are useful for change the cursor (grab/grabbing)

Is there another way to achieve this simply ?

NachoVazquez:
I agree those classes can become very handy, you could also increase the size of the element while grabbing to enhance the UX.

The draggable one is lest common to be used but I can see how someone could find use on that, like changing the cursor as @RectoVersoDev mentioned.

davidcetinkaya:
Thanks for the input guys. I don’t use this myself so I’m not opinionated about what should be included in this plugin and not. Do you want all these classes to have defaults or should they be empty unless you pass a value to the plugin?

RectoVersoDev:
Hello David,

1. Personally I don't have a preference, I let others say what they prefer

2. "ClassNames" seems good to me !

davidcetinkaya:
For anyone interested:

[...]

Issue #237, opened 2021-11-05
Plugins to create


davidcetinkaya:
# Feature Description
Create more plugins for Embla Carousel. Feel free to dig in. Don't wait for me to create all these.

### Suggestions of plugins to create:

- [x] [Autoplay](https://www.embla-carousel.com/plugins/autoplay/)
- [x] [Class Names](https://www.embla-carousel.com/plugins/class-names/)
- [x] [Auto Height](https://www.embla-carousel.com/plugins/auto-height/)
- [ ] Auto Scroll
- [ ] Fullscreen
- [ ] Fade
- [ ] Nav For
- [ ] Nested Carousels

davidcetinkaya:
This list has been transferred to discussions instead:

- #284

PR #242, opened 2021-11-22
Add Class Name plugin

davidcetinkaya:
# Add Class Name plugin

Solves the following issue:
- [x] #239


davidcetinkaya:
Partly solves: 
- #237

PR #243, opened 2021-11-23
Update wheel gestures plugin docs page

xiel:
- updates the contents for the wheel gestures plugin page with usage info

Thanks for including it. 🙏

### DIFF
--- a/packages/embla-carousel-docs/src/content/pages/plugins/wheel-gestures.md
+++ b/packages/embla-carousel-docs/src/content/pages/plugins/wheel-gestures.md
@@ -7,6 +7,37 @@ date: 2021-11-22
 
 # Wheel Gestures
 
-This plugin is used to extend Embla Carousel with...
+This plugin is used to extend Embla Carousel with the ability to use the mouse/trackpad wheel to navigate through the carousel.
 
-> ** NOTE! This section is still being written. **
+## Installation
+
+First you need to install the **npm package** and save it to your dependencies:
+
+```shell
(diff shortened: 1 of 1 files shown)


davidcetinkaya:
Thank you for this Felix 🚀. Out of curiosity: Do you provide an UMD bundle with the Wheel Gestures package? 

Best,
David

xiel:
Awesome, thanks @davidcetinkaya !

Currently there is no UMD bundle available, but I'll check how easy it is to make it available.

davidcetinkaya:
@xiel the UMD bundle is not a requirement so don’t sweat it if you have other important things to do 😃. I would say that it’s more nice to have.

Cheers!
David

xiel:
Was pretty easy to do, I sent you another very small PR. Makes it easier for users in the UMD world to use them all if needed.

davidcetinkaya:
@xiel just an idea: If you don’t want to update the Wheel Gestures documentation in two places in the future (it can be tedious), you can just add a link to the Embla Carousel docs page for the plugin in your Readme. 

Thanks for your efforts 🚀!

Best,
David

davidcetinkaya:
@xiel I also forgot to mention that I want to add a link to the GitHub repository and the plugin author on each plugin page.

xiel:
Sounds good 👏

davidcetinkaya:
[This feature](https://github.com/davidcetinkaya/embla-carousel/pull/243#issuecomment-983407266) will be added in:
- #255

PR #246, opened 2021-11-25
WheelGestures Plugin: Add CDN / UMD info to install section

xiel:
- WheelGestures Plugin is now also [available](https://github.com/xiel/embla-carousel-wheel-gestures/releases/tag/v2.0.2) as UMD bundle 📦


davidcetinkaya:
Nice work @xiel 🎉!
This is not a requirement for plugins so it's very nice of you to provide an UMD bundle.

Cheers,
David

PR #311, opened 2022-05-06
Docs: Add target option (Wheel Gestures Plugin)

xiel:
- add info about the newly added `target` option


davidjerleke:
Thank you for updating the docs @xiel ⭐.

PR #335, opened 2022-07-07
Typing of Embla Plugins

xiel:
- Implement Typing of EmblaPluginsType via [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) 
- Plugins extend the EmblaPluginsType (by interface merging) and we can keep the embla core independent of all external plugins
- This also ensures the types represented in the plugins() always match the types of the installed plugin version

Cool side effect, the typescript auto suggestion for e.g. `plugins().autoHeight` only appears once the plugin was installed & imported by the user.


davidjerleke:
Amazing @xiel ⭐!

This is exactly what I wanted to achieve and I didn't know it was possible with TypeScript. This approach is great because as you already mention:

> ...we can keep the embla core independent of all external plugins

With this problem out of the way we're very close to a stable v7 release 🙂. 

Massive thanks for this 🙏. Let me know if I can do anything for you.

Cheers,
David

PR #406, opened 2022-12-15
chore(docs): set custom root node for autoplay

openscript:


davidjerleke:
Thanks a bunch @openscript ❤️!

PR #575, opened 2023-09-22
Update slide-gaps.md

james0r:
Couldn't figure out how this worked until I chatbot'd it.


james0r:
Nevermind. This appears to add extra space when the last slide is in the viewport so i have no idea how to properly do this.

davidjerleke:
@james0r what do you want to achieve?

PR #586, opened 2023-09-29
Fix scrollTo signature on exported type

hill:
Hello!

The signature of `scrollTo` includes an optional `direction` parameter, however the type exposed by `EmblaCarouselType` excludes this parameter, leading me to extend the type definition in my own project in order to make use of it. It would be more convenient if it was exposed here in this type.

`scrollTo` signature:
```ts
scrollTo(index: number, jump?: boolean, direction?: number): void
```

Let me know if you have concerns about this and thanks for your work on Embla :)


davidjerleke:
Hi @hill,

Thanks for noticing 🙂. I left that out on purpose because it's used internally and I figured that devs will get confused about how it works:

- `-1` is actually forward because when the carousel scrolls forward, it translates it negatively. For a horizontal carousel the container with the slides moves to the left when it scrolls forward. And for vertical carousels, it moves upwards.
- `1` is backward.
- `0` means that Embla should choose the shortest way to its target whether it's backward of forward. 

We can add your type fix but if you also want to add it to the [`scrollTo()`](https://www.embla-carousel.com/api/methods/#scrollto) reference, please also take some time to explain [how it works here](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel-docs/src/content/pages/api/methods.mdx#L130-L135). Let me know if you want to do that.

Thank you for your contribution!

Best,
David

davidjerleke:
@hill let me know if you're interested in pursuing this when you've [read my response](https://github.com/davidjerleke/embla-carousel/pull/586#issuecomment-1740395306).

davidjerleke:
@hill last call. I will delete this if you don't respond soon.

PR #613, opened 2023-11-01
Fix leak in slidesInView event

fcasibu:
Hello first time trying out this package and I love everything about it but there's just one thing that's causing me some issues regarding the `slidesInView` event. I've created this [Sandbox](https://codesandbox.io/s/embla-carousel-autoplay-react-forked-p7jyy5?file=/src/js/EmblaCarousel.tsx) to demonstrate the issue.

To reproduce:
1. Try not to resize the sandbox and see the counter "called" gets incremented by 1 perfectly fine
2. Resize the sandbox
3. Wait for the autoplay to slide automatically or slide it yourself
4. Counter "called" gets incremented by A LOT and this shouldn't happen

I've added `engine.slidesInView.destroy()` in the `deactivate` function to fix the issue in `embla-carousel` package.

We can also listen to the `reinit` event to destroy the event but I feel like this shouldn't be done on the end user but on the package's side.

### DIFF
--- a/packages/embla-carousel/src/components/EmblaCarousel.ts
+++ b/packages/embla-carousel/src/components/EmblaCarousel.ts
@@ -155,6 +155,7 @@ function EmblaCarousel(
     engine.slideLooper.clear()
     engine.resizeHandler.destroy()
     engine.slidesHandler.destroy()
+    engine.slidesInView.destroy()
     pluginsHandler.destroy()
     mediaHandlers.clear()
     documentVisibleHandler.clear()


davidjerleke:
@fcasibu thank you very much for your effort! Have you tried the changes to see if it solves the problem?

Best,
David

fcasibu:
Hey @davidjerleke I've tested it on the playground and it does indeed solve the problem. Thanks 🙇

davidjerleke:
@fcasibu thank you for your swift response. Nice work and welcome to the [contributors club](https://github.com/davidjerleke/embla-carousel#contributors) 🎉!

fcasibu:
Seems like it broke a lot of tests regarding `intersectionObserver` should've added `disconnect` on the mock, sorry about that @davidjerleke

davidjerleke:
@fcasibu no worries. It's just a minor thing. Thanks for informing me though 👍.

Issue #647, opened 2023-12-07
`EmblaOptionsType` doesn't exists in React


proninyaroslav:
### Bug is related to

- [ ] embla-carousel (core package)
- [x] embla-carousel-react
- [x] embla-carousel-vue
- [x] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-docs (documentation)
- [ ] embla-carousel-docs (generator)

### Embla Carousel version

- `v8.0.0-rc15`

### Describe the bug

Since version v8.0.0-rc15,  ESLint says that the `SliderPropsOptions` type does not have a `loop` property. This happens to all properties that are in `EmblaOptionsType`:

![image](https://github.com/davidjerleke/embla-carousel/assets/7840559/420a01d6-a92b-4d8a-b9af-c1efea9fb98f)

Also because of this, `useEmblaCarousel` has the `options` argument type `Partial<OptionsType> | undefined` instead of `EmblaOptionsType | undefined`, which causes a linter error:

![image](https://github.com/davidjerleke/embla-carousel/assets/7840559/0463401c-d806-421a-a9e5-5a662fefbc35)

davidjerleke:
Thanks @proninyaroslav,

I’ve noted this bug and already started working on it. However, thanks for creating a bug report so I can connect it to a pull request and devs can track it 👍🏻.

Best,
David

davidjerleke:
@proninyaroslav a bug fix for this was just released in [`v8.0.0-rc16`](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.0.0-rc16). Try it out and let me know if it's working as expected.

proninyaroslav:
@davidjerleke 
Updated to v8.0.0-rc17 and still have this problem.

davidjerleke:
@proninyaroslav then I suggest you create a CodeSandbox with your setup in order to demonstrate the problem. You didn’t provide one in your bug report but it’s a requirement as mentioned in the contribution guidelines.

davidjerleke:
@proninyaroslav I think it's better to get the types directly from the main package. Because `embla-carousel` is a dependency, you can do this:

```tsx
import { EmblaOptionsType } from 'embla-carousel'; // Get it like so
import useEmblaCarousel from 'embla-carousel-react';
```

[...]

PR #658, opened 2023-12-20
Add TypeScript usage examples to the docs

zaaakher:
Added typescript page for the docs

--- 

This is just a starting point since I had to wing it since I couldn't get the gatspy docs running locally in my PC. I will add more when I get the chance god willing

### DIFF
--- a/packages/embla-carousel-docs/gatsby-node.ts
+++ b/packages/embla-carousel-docs/gatsby-node.ts
@@ -2,6 +2,7 @@ import path from 'path'
 import { GatsbyNode } from 'gatsby'
 import { createFilePath } from 'gatsby-source-filesystem'
 import { GraphQLAllDataType, GraphQLNodeType } from './src/consts/graphQL'
+import { PRISM_HIGHLIGHT_PROP_SEPARATOR } from './src/consts/prismHighlight'
 import {
   PAGE_LAYOUTS,
   PageOverviewType,
@@ -14,6 +15,8 @@ import {
 } from './src/consts/page'
 
 const PAGE_LAYOUTS_ROOT = './src/templates'
+const CODE_BLOCK_START_REGEX = /```.*/g
+const EMPTY_SPACE_REGEX = /\s/g
 
 const PAGE_TEMPLATES = {
   HOME: path.resolve(`${PAGE_LAYOUTS_ROOT}/${PAGE_LAYOUTS.HOME}.tsx`),
@@ -66,12 +69,21 @@ const createPageListWithChildren = (
     )
 }
 
+const addPrismCodeBlockPropsSupport = (body: string): string => {
+  return body.replace(CODE_BLOCK_START_REGEX, (match) =>
+    match.replace(EMPTY_SPACE_REGEX, PRISM_HIGHLIGHT_PROP_SEPARATOR)
+  )
+}
+
 export const onCreateNode: GatsbyNode['onCreateNode'] = ({
   node,
   actions: { createNodeField },
   getNode
 }) => {
   if (node.internal.type === 'Mdx') {
+    const body = <string>node.body
+    node.body = addPrismCodeBlockPropsSupport(body)
+
     const value = createFilePath({ node, getNode })
     createNodeField({ node, value, name: 'slug' })
   }
--- a/packages/embla-carousel-docs/src/components/CarouselGenerator/CarouselGeneratorAlignmentSettings.tsx
+++ b/packages/embla-carousel-docs/src/components/CarouselGenerator/CarouselGeneratorAlignmentSettings.tsx
@@ -4,7 +4,7 @@ import CarouselDefault from 'components/Sandbox/React/SandboxFilesSrc/Default/Em
 import { InputRadioDefault } from 'components/Input/InputRadio'
 import { InputCheckboxDefault } from 'components/Input/InputCheckbox'
 import { arrayFromNumber } from 'utils/arrayFromNumber'
-import { EmblaOptionsType } from 'embla-carousel-react'
+import { EmblaOptionsType } from 'embla-carousel'
 import { createCarouselDefaultStyles } from 'components/Examples/createCarouselStyles'
 import { Admonition } from 'components/Mdx/Components/Admonition'
 import { BRAND_GRADIENT_BACKGROUND_STYLES } from 'consts/gradients'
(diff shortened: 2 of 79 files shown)


davidjerleke:
Note: We need to [add this](https://github.com/davidjerleke/embla-carousel/issues/647#issuecomment-1868273239) to the docs.

zaaakher:
@davidjerleke I finalized the typescript page and also updated the way types are imported in the Sandboxes.

Kindly take a look at it and let me know you need additional help 👍

zaaakher:
Hey @davidjerleke ,

I've been working on my UI kit library [here](https://github.com/sikka-software/hawa) and I have the carousel component working well with only using `embla-carousel-react` package. Now that the types will move into `embla-carousel` package, does that mean I have to install it just to use the types? would it be bad to install `embla-carousel` as a devDependency? 

Or should we look into making a PR for creating a `@types/embla-carousel` package in [this repo](https://github.com/DefinitelyTyped/DefinitelyTyped)?

Or would it be strategic to make a package solely for the types? like `embla-carousel-types` ?

Based on your comment [here](https://github.com/davidjerleke/embla-carousel/issues/647#issuecomment-1868273239) , `embla-carousel` should be a dependency but somehow I got the carousel working without `embla-carousel` and with only the react wrapper (here's my [package.json](https://github.com/sikka-software/Hawa/blob/a13896076dc51241df9ce3f464a2fc4a0b23e36a/package.json#L89))

davidjerleke:
@zaaakher `embla-carousel-react` has `embla-carousel` as a dependency, so when you install the react package it will automatically install the core package under the hood. You can easily confirm this like so:

- Install `embla-carousel-react`.
- Expand the `node_modules` folder.
- Scroll down to `e` and see both embla packages there: `embla-carousel-react` and `embla-carousel`.

So you should import types from the core package `embla-carousel`. I will remove any type re-exports from all library/framework packages in the next release.

I hope my explanation makes sense?

Best,
David

zaaakher:
> I don’t think it’s common practice to make a separate types package because Embla has first class TypeScript support. It’s built with TypeScript. I think type packages are mainly used for packages built without TypeScript.

True true, I agree.

---

> `embla-carousel-react` has `embla-carousel` as a dependency, so when you install the react package it will automatically install the core package under the hood

That doesn't seem to be working as expected when using pnpm workspace

in my project [here](https://github.com/sikka-software/Hawa) I can do as you mentioned and everything works fine.

But currently i'm migrating to a proper monorepo structure using pnpm (in the [monorepo branch](https://github.com/sikka-software/Hawa/tree/monorepo)) and it seems `embla-carousel` is not recognized.

image

zaaakher:
Looking at [this](https://github.com/pnpm/pnpm/issues/2903) I was able to find `embla-carousel` in the root `node_modules` inside `.pnpm` but doing the following won't work and doesn't even look right.
```
import { EmblaOptionsType } from "../../../../node_modules/.pnpm/embla-carousel@8.0.0-rc17";
```

It's most likely a `pnpm` thing but I think it could be an important caveat to include somewhere in this library.

zaaakher:
shadcn-ui is using this library but they still [didn't migrate](https://github.com/search?q=repo%3Ashadcn-ui%2Fui%20embla-carousel&type=code) to the new typescript approach. And they're using `pnpm` as well. 

So there's a chance they might stumble upon a similar problem if not an identical one to mine as they update embla as per your suggestion [here](https://github.com/shadcn-ui/ui/discussions/2146) and change the way they import types. 

Of course a simple solution would be to explicitly install `embla-carousel` but I'm inclined to resist to do that.

davidjerleke:
@zaaakher that’s valuable information. Thank you. I haven’t used `pnpm` so didn’t know about this. If this is the case, the solution would be to copy over the types from the core package to the library wrappers during the build process. Are there anymore things to consider to fully support `pnpm`?

zaaakher:
> Are there anymore things to consider to fully support pnpm?

Making `embla-carousel` a peer dependency of `embla-carousel-react` could potentially solve this problem (I'm not entirely sure if it will). But would doing that cause problems elsewhere?

davidjerleke:
> Making `embla-carousel` a peer dependency of `embla-carousel-react` could potentially solve this problem (I'm not entirely sure if it will). But would doing that cause problems elsewhere?

@zaaakher I’m not sure how that would affect npm/yarn and pnpm for that matter…

zaaakher:
> Is that a conventional solution?

It doesn't feel like it and i'm very doubtful if it'd work.

> the solution would be to copy over the types from the core package to the library wrappers during the build process

That's a good solution, I can't think of a better one at the moment. Although is there a possible scenario in the future where a library wrapper uses a slightly different type from the one in the core library? (i.e. something unique to react or vue or server components) 

Currently I'm using types the original way 👇
```jsx
import useEmblaCarousel, { EmblaOptionsType } from "embla-carousel-react";
```
I think it could be a good developer experience when consumers of this library have to install a single package (depending on their framework) where they can import hooks, types, and possibly components. 

Honestly I never knew I could import from a package that's not listed in my package.json, I just learned I could do that from you 😄 (although that didn't work with pnpm).

I wish I have more time on my hand to test it out with pnpm alternatives like [lerna](https://lerna.js.org/), [rush](https://rushjs.io/), or [bolt](https://github.com/boltpkg/bolt). Nevertheless, I have a feeling they all have a similar approach to how they put aside dependencies-of-dependencies to save disk space and increase performance within a monorepo structure.

I understand there could be performance/size trade-offs depending on the approach you'd like to take and I'm here if you wanna bounce ideas around, I'm just thinking out loud.

davidjerleke:
@zaaakher I will try the copy types approach then. Thanks for your thoughts 👍🏻.

Wow. The frontend tooling world is so broken. It’s really a nightmare to try to support everything. Feels like we spend 95% of our time with tools…

zaaakher:
> Wow. The frontend tooling world is so broken. It’s really a nightmare to try to support everything. Feels like we spend 95% of our time with tools…

True. It's the same cycle where an individual or a group of people make tool, then it gains popularity, then it's a either they have to start adapting to others or others have to adapt to them.

mulfyx:
Workaround:

```ts
import { UseEmblaCarouselType } from "embla-carousel-react";

type EmblaCarouselType = NonNullable<UseEmblaCarouselType[1]>;
```

davidjerleke:
@zaaakher again, thanks for all your contributions and making this library better 🌟!

zaaakher:
@davidjerleke You're very welcome brother 🌹

mulfyx:
i think you miss something

image

zaaakher:
Yup @mulfyx , it'll be fixed with #685

PR #673, opened 2023-12-31
Use `onBeforeUnmount` instead of `onUnmounted` to separate embla DOM cleanup from Vue Reactivity

sadeghbarati:


davidjerleke:
Thank you for your contribution @sadeghbarati!

PR #683, opened 2024-01-07
Development walkthrough in CONTRIBUTING.md

zaaakher:
Once again, thanks @davidjerleke for creating this wonderful and promising library ✨

My goal in this PR is to thoroughly detail this project in a way that makes it less intimidating for developers who want to contribute to the inner workings of this project. I will continue adding to it every time I get a chance and I will hopefully learn more about the inner workings of this project as I expand on this contribution guide. 

@davidjerleke feel free to throw in a commit every now and then to fix things or add sections or even notes that I can tag-team and expand on.

The good thing about MDX is that if this contribution guide becomes too big we can always move it to `embla-carouse-docs` and break the guide down into multiple sub-pages.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,63 @@
-## Contributing
+# Contributing
 
 Thank you for considering contributing to Embla Carousel, contributions are welcome! When contributing, your code will be used as part of an open source product if merged. By submitting a Pull Request, **you are giving your permission** for your code to be integrated into Embla Carousel as part of an open source product.
 
+- [Development](#development)
 - [Bug reports](#bug-reports)
 - [New features](#new-features)
 - [Documentation](#documentation)
 - [Code style](#code-style)
 
-### Bug reports
+## Development
+
+This project is a monorepo managed by `yarn` workspaces. It's highly recommended to use `yarn` to take advantage of the `yarn.lock` file to keep your dependencies in alignment with the maintainers of this project.
+
+### Prerequisites
+
+- **NodeJS** version equal to or higher than the one indicated in [.nvmrc](https://github.com/davidjerleke/embla-carousel/blob/master/.nvmrc).
+- `yarn`
+
+### Structure
+
+#### Packages
+
+These are the actual published packages which can be installed from the npm registry.
+
+| Purpose | Package                                                                                                                            | Size                                                                                                                                                                                                        | Version                                                                                                                                                              |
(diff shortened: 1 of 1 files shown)


zaaakher:
I forgot to remove the commented part, I was separate the bullet-points by the categories of the packages but then realized it makes more sense to list them the same way they're grouped in the folders of this monorepo.

Now that I think about it I think it would be better to make them into a table with a badge next to each package that dynamically shows the size and version (same as the badges of each README.md inside each package)

davidjerleke:
> I was thinking about a **versioning** section inspired from your comment [here](https://github.com/davidjerleke/embla-carousel/pull/685#issuecomment-1880405557) but on second thoughts I think that's something maybe contributors shouldn't touch since you might have a specific approach to putting together PRs and deciding when to publish.

This is the publishing process for the packages. Running any of these commands below:

https://github.com/davidjerleke/embla-carousel/blob/01036845655e843e21eac50f28c0e61dab06ed4c/package.json#L51-L53

...will automatically:

- Bump the versions of all `Embla` packages in the entire repository.
- Update the `yarn.lock` file if needed.
- Add all file changes and a new tag with the new version to git.
- Push the tag and commit to the remote repository.

The publishing part is the only thing that needs manual work. In order to publish I have to manually create a new release on GitHub with the new tag version and hit `create release`. After that the CD process takes over and automatically publishes all packages to the `npm` registry **if** `build`, `test` and `lint` pass.

But at the time of writing, I guess I'm the only one that have the access rights to creating a release.

Best,
David

zaaakher:
Sounds great 👍 

Although I was wondering more about **when** a contributor should bump a version. So for example if I make a PR fixing a few lines of codes in one of the packages, would you prefer/recommend I bump a patch version in my commit? Or would that disrupt your flow and you could do a bump as well (without noticing my bump) and we end up with one of the packages version not in-sync with the others?

I could be wrong, but you sometimes do 1 bump for multiple PR's you combine, right?

Overall, if you prefer contributors include do some versioning in PR's then I think we should include it in the `CONTRIBUTING.md` otherwise, we can just leave out the versioning section.

davidjerleke:
@zaaakher sorry I misunderstood your question:

> I could be wrong, but you sometimes do 1 bump for multiple PR's you combine, right?

That's correct. A new version can include multiple **bug fixes** or **features**.

> Overall, if you prefer contributors include do some versioning in PR's then I think we should include it in the `CONTRIBUTING.md` otherwise, we can just leave out the versioning section.

I would prefer not at this stage but very much appreciate you asking 🙂!

davidjerleke:
@zaaakher great work as always 🥳.

PR #685, opened 2024-01-08
Added `embla-carousel` as a dependency for CodeSandbox template

### LINKED ISSUE #684: Types warning in (TS + React) CodeSandbox template
I hate to be the bearer of bad news but codesandbox **might** be using pnpm under the hood 😅

### Bug is related to

- [x] embla-carousel-docs

### Embla Carousel version

- `8.0.0-rc19`

### Describe the bug

- The generated **(React + TS)** codesandbox shows a warning it's not recognizing `embla-carousel` (see image below)

image

### Steps to reproduce

1. Go to the [predefined examples page](https://www.embla-carousel.com/examples/predefined/)
2. Click on "edit code" in any of the examples
3. Choose "**React + Typescript**" and create the sandbox
4. Open a file that's importing a type from `embla-carousel` package
5. See the TS warning like in the image above

### Expected behavior

- No TS warnings should exist.

zaaakher:
Fixes #684

I had to add `embla-carousel` as a dependency of `embla-carousel-docs` so that it can be used in `createSandboxReactPackageJson.ts` dynamically

### DIFF
--- a/packages/embla-carousel-docs/package.json
+++ b/packages/embla-carousel-docs/package.json
@@ -38,6 +38,7 @@
     "@mdx-js/react": "^2.2.1",
     "babel-plugin-styled-components": "^2.0.2",
     "codesandbox": "^2.2.3",
+    "embla-carousel": "8.0.0-rc19",
     "embla-carousel-autoplay": "8.0.0-rc19",
     "embla-carousel-class-names": "8.0.0-rc19",
     "embla-carousel-react": "8.0.0-rc19",
--- a/packages/embla-carousel-docs/src/components/Sandbox/React/createSandboxReactPackageJson.ts
+++ b/packages/embla-carousel-docs/src/components/Sandbox/React/createSandboxReactPackageJson.ts
@@ -34,6 +34,7 @@ export const createSandboxReactPackageJson = async (
       'react-dom': dependencies['react-dom'],
       'react-scripts': '4.0.0',
       'embla-carousel-react': dependencies['embla-carousel-react'],
+      'embla-carousel': dependencies['embla-carousel'],
       ...(plugins && plugins)
     },
     devDependencies: isJavaScript
--- a/yarn.lock
+++ b/yarn.lock
@@ -8804,6 +8804,7 @@ __metadata:
     "@typescript-eslint/parser": ^6.9.0
     babel-plugin-styled-components: ^2.0.2
     codesandbox: ^2.2.3
+    embla-carousel: 8.0.0-rc19
     embla-carousel-autoplay: 8.0.0-rc19
     embla-carousel-class-names: 8.0.0-rc19
     embla-carousel-react: 8.0.0-rc19


zaaakher:
I might be missing something, 

What do you use to update the embla-carousel versions in all the package.json's of all the packages in this monorepo? and is it going to automatically update this newly added `embla-carousel` to `embla-carousel-docs` package.json?

packages/embla-carousel-docs/package.json
```diff
@@ -41,6 +41,7 @@
     "embla-carousel-autoplay": "8.0.0-rc19",
     "embla-carousel-class-names": "8.0.0-rc19",
     "embla-carousel-react": "8.0.0-rc19",
+    "embla-carousel": "8.0.0-rc19",
```
zaaakher, line 44: Is this going to be automatically updated as part of the automation process you use? Or do we have to make a change somewhere else to include it in the update-versions-automation?

davidjerleke:
> I might be missing something,
> 
> What do you use to automate updating the embla-carousel** versions in all the package.json's of all the packages in this monorepo?

@zaaakher good question! These commands in the root `package.json` will do that:

https://github.com/davidjerleke/embla-carousel/blob/01036845655e843e21eac50f28c0e61dab06ed4c/package.json#L51-L53

Best,
David

packages/embla-carousel-docs/package.json
davidjerleke: Yarn solves this automatically 🙂. But shouldn’t we add the core package to `devDependencies` instead? Thanks as always!

zaaakher: >But shouldn’t we add the core package to devDependencies instead
Yup you're right, I just pushed a commit fixing that 👍

zaaakher:
@davidjerleke so it seems that the unrecognized `embla-carousel` in CodeSandbox is not resolved yet.

 The only way I was able to fix it is by making `embla-carousel` a dependency in `createSandboxReactPackageJson.ts` like this:
```jsx
    dependencies: {
      react: dependencies.react,
      'react-dom': dependencies['react-dom'],
      'react-scripts': '4.0.0',
      'embla-carousel-react': dependencies['embla-carousel-react'],
      'embla-carousel': dependencies['embla-carousel'], <--- here (assuming embla-carousel moved back to dependencies)
      ...(plugins && plugins)
    },
```


Does that mean I should move `embla-carousel` back as a regular dependency in `embla-carousel-docs` like before?

davidjerleke:
> @davidjerleke so it seems that the unrecognized `embla-carousel` in CodeSandbox is not resolved yet.
> 
> The only way I was able to fix it is by making `embla-carousel` a dependency in `createSandboxReactPackageJson.ts` like this:
> 
> ```js
>     dependencies: {
>       react: dependencies.react,
>       'react-dom': dependencies['react-dom'],
>       'react-scripts': '4.0.0',
>       'embla-carousel-react': dependencies['embla-carousel-react'],
>       'embla-carousel': dependencies['embla-carousel'], <--- here (assuming embla-carousel moved back to dependencies)
>       ...(plugins && plugins)
>     },
> ```
> 
> Does that mean I should move `embla-carousel` back as a regular dependency in `embla-carousel-docs` like before?

If I spin up a new `npm` or `yarn` project locally and import the types from `embla-carousel` when using `embla-carousel-react`, it just works. And with `pnpm` installing it as a `devDependency` works. Maybe CodeSandbox handles dependencies differently than all package managers? 

**So yes:** If adding it as a `dependency` works, we have to settle with this special solution for the CodeSandboxes only. Thank you 👍!

zaaakher:
Don't worry about it 👌 

I just pushed a commit and everything works properly now 👍

PR #729, opened 2024-02-01
Update 'events' Typescript example useEffect deps

ranno-lauri:
The example is never run, if emblaApi is not added as a dependancy to the useEffect


davidjerleke:
Thanks a lot @ranno-lauri for fixing this. Welcome to the contributors club 🌟!

PR #805, opened 2024-03-26
bug fix: render when slides are wider than view size

tolu-fuelled:
Hi David,

Thanks for a great library. You efforts are much appreciated.

This PR addresses an issue I ran into - when the first slide is wider than the view size, an exception is thrown. This happens because of an underlying assumption when creating scroll groups by size (see test case).

Happy to make additional changes if any suggestions.

### DIFF
--- a/packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
+++ b/packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
@@ -8,6 +8,47 @@ import {
 
 const FIRST_SNAP_INDEX = 0
 
+describe('➡️  SlidesToScroll - Horizontal LTR - Wide slides', () => {
+  describe('"auto" is correct for wide slides WITHOUT MARGINS and:', () => {
+    const emblaApi = EmblaCarousel(
+      mockTestElements({
+        ...FIXTURE_SLIDES_TO_SCROLL_LTR_1,
+        containerOffset: {
+          offsetWidth: 250,
+          offsetHeight: 190,
+          offsetTop: 0,
+          offsetLeft: 0
+        }
+      })
+    )
+
+    beforeEach(() => {
+      emblaApi.reInit({ ...defaultOptions, slidesToScroll: 'auto' })
+    })
+
+    test('LOOP:FALSE', () => {
+      const engine = emblaApi.internalEngine()
+      const expectedScrollSnaps = [
+        0, -625, -1000, -1250, -1500, -1750, -2125, -2625, -3125, -3751
+      ]
+      expect(engine.scrollSnaps).toEqual(expectedScrollSnaps)
+      expect(engine.location.get()).toBe(expectedScrollSnaps[FIRST_SNAP_INDEX])
+      expect(engine.slideRegistry).toEqual([
+        [0],
+        [1],
+        [2],
+        [3],
+        [4],
+        [5],
+        [6],
+        [7],
+        [8],
+        [9]
+      ])
+    })
+  })
+})
+
 describe('➡️  SlidesToScroll - Horizontal LTR', () => {
   describe('"auto" is correct for slides WITHOUT MARGINS and:', () => {
     const emblaApi = EmblaCarousel(
--- a/packages/embla-carousel/src/components/SlidesToScroll.ts
+++ b/packages/embla-carousel/src/components/SlidesToScroll.ts
@@ -38,7 +38,7 @@ export function SlidesToScroll(
     if (!array.length) return []
 
     return arrayKeys(array)
-      .reduce((groups: number[], rectB) => {
+      .reduce((groups: number[], rectB, index) => {
         const rectA = arrayLast(groups) || 0
         const isFirst = rectA === 0
         const isLast = rectB === arrayLastIndex(array)
@@ -49,7 +49,8 @@ export function SlidesToScroll(
         const gapB = !loop && isLast ? direction(endGap) : 0
         const chunkSize = mathAbs(edgeB - gapB - (edgeA + gapA))
 
-        if (chunkSize > viewSize + pixelTolerance) groups.push(rectB)
+        if (index !== 0 && chunkSize > viewSize + pixelTolerance)
+          groups.push(rectB)
         if (isLast) groups.push(array.length)
         return groups
       }, [])


davidjerleke:
@tolu-fuelled thanks a lot for your contribution ✨! I will check this out as soon as possible.

One thing I noticed (that isn’t easy to know): We can get rid of the last commit with the patch version bumps on this bug fix branch, because this will be done automatically on the main branch by running `yarn version:patch` - which automatically creates a release and pushes it to the remote. The next patch version `8.0.1` will include several bug fixes.

Thanks a lot and I’ll let you know when I’ve reviewed this.

Best,
David

suriyaprakash-s:
This is awesome, thanks for the fix @tolu-fuelled I have been waiting for this for long time. @davidjerleke When can we expect this to be officially released?

davidjerleke:
@suriyaprakash-s I can never give time estimates because I work on this library on my unpaid spare time.

However, reviewing this PR is next up after a minor release with some other bug fixes that has higher priority.

packages/embla-carousel/src/components/SlidesToScroll.ts
```diff
@@ -49,7 +49,8 @@ export function SlidesToScroll(
         const gapB = !loop && isLast ? direction(endGap) : 0
         const chunkSize = mathAbs(edgeB - gapB - (edgeA + gapA))
 
-        if (chunkSize > viewSize + pixelTolerance) groups.push(rectB)
+        if (index !== 0 && chunkSize > viewSize + pixelTolerance)
```
davidjerleke, line 52: We can modify this like so:

```suggestion
        if (index && chunkSize > viewSize + pixelTolerance)
```

packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
```diff
@@ -8,6 +8,47 @@ import {
 
 const FIRST_SNAP_INDEX = 0
 
+describe('➡️  SlidesToScroll - Horizontal LTR - Wide slides', () => {
+  describe('"auto" is correct for wide slides WITHOUT MARGINS and:', () => {
```
davidjerleke, line 12: Let's move this test to `line 108` in this file **after** all other related tests and rephrase it so it's clear that we're testing:

- The case when **slides are wider** than the **viewport**.
- An **edge case**. See existing example from a different test file:
https://github.com/davidjerleke/embla-carousel/blob/1588deacf392c8ace7cb3f243396e769d1fe853a/packages/embla-carousel/src/__tests__/containScroll-ltr.test.ts#L459-L460

packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
```diff
@@ -8,6 +8,47 @@ import {
...
+          offsetHeight: 190,
+          offsetTop: 0,
+          offsetLeft: 0
+        }
+      })
+    )
+
+    beforeEach(() => {
+      emblaApi.reInit({ ...defaultOptions, slidesToScroll: 'auto' })
+    })
```
davidjerleke, line 27: We can get rid of this then because this `beforeEach` already exists in the closure that I suggested that we move this edge case test to.

packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
```diff
@@ -8,6 +8,47 @@ import {
...
+describe('➡️  SlidesToScroll - Horizontal LTR - Wide slides', () => {
+  describe('"auto" is correct for wide slides WITHOUT MARGINS and:', () => {
+    const emblaApi = EmblaCarousel(
+      mockTestElements({
+        ...FIXTURE_SLIDES_TO_SCROLL_LTR_1,
+        containerOffset: {
+          offsetWidth: 250,
+          offsetHeight: 190,
+          offsetTop: 0,
+          offsetLeft: 0
```
davidjerleke, line 20: Let's create a **new fixture** [here](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/__tests__/fixtures/slidesToScroll-ltr.fixture.ts#L165) for this specific case instead of mixing other fixtures with local overrides.

davidjerleke (changes requested):
Very nice work and thanks for contributing ⭐!

- In addition to some minor change requests from my side, I would like to see the same test case being added for `RTL` and `Vertical` setups.

- I wanted to rebase your branch because I've merged some stuff but it seems like I don't have access rights to do so. Maybe because you didn't create a feature branch for this and made the changes on your forked repo `master` branch? I don't know. Could you rebase it so it's up to date with my `master` branch?

packages/embla-carousel/src/__tests__/slidesToScroll-ltr.test.ts
tolu-fuelled: @davidjerleke given that we need to initialize `emblaApi` with a different fixture I'm missing how we get rid of this `beforeEach` block.

davidjerleke: @tolu-fuelled sorry. I was a bit too fast there. You’re right 🙂.

davidjerleke: @tolu-fuelled not to rush you, but did you see [my comment](https://github.com/davidjerleke/embla-carousel/pull/805#discussion_r1554648047)?

tolu-fuelled: @davidjerleke I did - was a busy weekend. Should be able to submit the changes today.

tolu-fuelled:
@davidjerleke made the requested changes and added a few more tests.

The RTL snaps don't add up to me if you don't mind giving it a second look. I ran the react playground - the slides render left to right but they move right to left. So slides render [1, 2, 3, 4, 5] but move [1, empty, empty, empty, empty]. I'm probably missing something.

davidjerleke:
> @davidjerleke made the requested changes and added a few more tests.

Thanks! It looks great 👍. 

> The RTL snaps don't add up to me if you don't mind giving it a second look. I ran the react playground - the slides render left to right but they move right to left. So slides render [1, 2, 3, 4, 5] but move [1, empty, empty, empty, empty]. I'm probably missing something.

@tolu-fuelled you've set the [direction option](https://www.embla-carousel.com/api/options/#direction) to `'rtl'` but my guess would be that you've forgot to set the content direction to RTL too, because Embla respects the native browser direction and doesn't shift around your slides to achieve a RTL carousel. [From the docs](https://www.embla-carousel.com/api/options/#direction):

> [!NOTE]  
> Note: When using `rtl`, the content direction also has to be set to RTL, either by using the [HTML dir attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) or the [CSS direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction) property.

Could that be the case?

Best,
David

tolu-fuelled:
@davidjerleke that’s exactly it.

El 9 abr 2024, a la(s) 00:57, David Jerleke ***@***.***> escribió:

﻿

@davidjerleke<https://github.com/davidjerleke> made the requested changes and added a few more tests.

Thanks! It looks great 👍.

The RTL snaps don't add up to me if you don't mind giving it a second look. I ran the react playground - the slides render left to right but they move right to left. So slides render [1, 2, 3, 4, 5] but move [1, empty, empty, empty, empty]. I'm probably missing something.

@tolu-fuelled<https://github.com/tolu-fuelled> you've set the direction option<https://www.embla-carousel.com/api/options/#direction> to 'rtl' but my guess would be that you've forgot to set the content direction to RTL too, because Embla respects the native browser direction and doesn't shift around your slides to achieve a RTL carousel. From the docs<https://www.embla-carousel.com/api/options/#direction>:

Note

Note: When using rtl, the content direction also has to be set to RTL, either by using the HTML dir attribute<https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir> or the CSS direction<https://developer.mozilla.org/en-US/docs/Web/CSS/direction> property.

Could that be the case?

Best,
David

davidjerleke (approved):

davidjerleke:
@tolu-fuelled great work! Thanks for solving this problem and stabilizing Embla with relevant tests. It will prevent the bug from reoccuring in the future 👍.

davidjerleke:
@tolu-fuelled this has been released in [v8.0.2](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.0.2).

PR #830, opened 2024-04-19
[Feat]: Add robust pinch zoom support

realwagyu:
- Closes: #828.

### DIFF
--- a/packages/embla-carousel/src/components/DragHandler.ts
+++ b/packages/embla-carousel/src/components/DragHandler.ts
@@ -147,6 +147,8 @@ export function DragHandler(
   }
 
   function move(evt: PointerEventType): void {
+    const isTouchEvt = !isMouseEvent(evt, ownerWindow)
+    if (isTouchEvt && evt.touches.length >= 2) return up(evt)
     const lastScroll = dragTracker.readPoint(evt)
     const lastCross = dragTracker.readPoint(evt, crossAxis)
     const diffScroll = deltaAbs(lastScroll, startScroll)
--- a/packages/embla-carousel-docs/src/components/Examples/examplesCarouselStyles.ts
+++ b/packages/embla-carousel-docs/src/components/Examples/examplesCarouselStyles.ts
@@ -635,7 +635,7 @@ export const examplesCarouselDefaultStyles = (
   const horizontal = axis === 'x'
   const flexDirection = horizontal ? '' : 'flex-direction: column;'
   const spacingDirection = horizontal ? 'left' : 'top'
-  const panDirection = `pan-${horizontal ? 'y' : 'x'}`
+  const panDirection = `pan-${horizontal ? 'y' : 'x'} pinch-zoom`
   const sizeDimention = horizontal ? 'width' : 'height'
   const slideHeight = horizontal ? 'var(--slide-height)' : '100%'
   const containerHeight = horizontal


davidjerleke:
@realwagyu thank you for your contribution 👍. I will review this when possible. Thanks for your patience.

Best,
David

packages/embla-carousel/src/components/DragHandler.ts
```diff
@@ -147,6 +147,9 @@ export function DragHandler(
   }
 
   function move(evt: PointerEventType): void {
+    if (window.TouchEvent && evt instanceof TouchEvent && evt.touches.length >= 2) {
+      return;
+    }
```
davidjerleke, line 152: Maybe we should release any ongoing scroll when there are multiple touches? Additionally there's already a utility function for checking if the `event` is a `TouchEvent` or a `MouseEvent`. It makes use of the `ownerWindow` instead of `window` which supports carousels from other realms like iframes.

So we could change your suggestion to this:

```suggestion
    const isTouchEvt = !isMouseEvent(evt, ownerWindow)
    if (isTouchEvt && evt.touches.length >= 2) return up(evt)
```

<br/>

I've tested on Safari and it behaves like this:


- Pinch zoom works in both directions (both the scroll direction and the opposite).
- A pinch move doesn't scroll the carousel (it terminates any ongoing `touchmove` listeners).
- If only one finger is touching the screen initially and it scrolls the carousel, and you add an additional finger (2 total), the carousel stops scrolling, but you have to release your fingers and perform a new pinch move to zoom.

davidjerleke (changes requested):
Great initiative 👍!

I have one change request and would very much appreciate if you could test it on an Android device.

Thanks in advance!

packages/embla-carousel/src/components/DragHandler.ts
realwagyu: Yes, great!
With the modified commit, it works correctly/as you describe on:
- Android Chromium 116
- Android Firefox 124
- Android Opera Mini 79

davidjerleke: @realwagyu thanks for testing! Feel free to push the modification and the additional change suggested [here](https://github.com/davidjerleke/embla-carousel/pull/830#pullrequestreview-2026630952) (see point 2) when you want. 

Thanks!

realwagyu: Amended with both changes: https://github.com/davidjerleke/embla-carousel/pull/830/commits/6f5a6c96b27893ffe2a2186f100a3641e3054139

davidjerleke: @realwagyu thanks a ton. I will merge this soon 🎉.

davidjerleke (approved):

packages/embla-carousel/src/components/DragHandler.ts
realwagyu: Thank you for this beautiful library

davidjerleke: @realwagyu thanks! Would you mind rebasing your [commit](https://github.com/davidjerleke/embla-carousel/pull/830/commits/8d5eb22c00d95a98fbd0a7d774d075460c4f06b8) on the latest `davidjerleke/embla-carousel:master` and I will merge this afterwards?

realwagyu: @davidjerleke done, let me know if you need anything else

PR #836, opened 2024-04-22
chore(docs): note about reactive options

meirroth:
This PR adds a note mentioning that in some Embla wrappers it's possible to pass reactive options and Embla will automatically reinitialize when they change.

### DIFF
--- a/packages/embla-carousel-docs/src/content/pages/api/events.mdx
+++ b/packages/embla-carousel-docs/src/content/pages/api/events.mdx
@@ -66,7 +66,7 @@ export function EmblaCarousel() {
 
 ```html highlight={12}
 <script setup>
-  import { watchEffect } from 'vue'
+  import { onMounted } from 'vue'
   import emblaCarouselVue from 'embla-carousel-vue'
 
   const [emblaRef, emblaApi] = emblaCarouselVue()
@@ -75,7 +75,7 @@ export function EmblaCarousel() {
     console.log(emblaApi.slidesInView())
   }
 
-  watchEffect(() => {
+  onMounted(() => {
     if (emblaApi.value) emblaApi.value.on('slidesInView', logSlidesInView)
   })
 
@@ -87,7 +87,7 @@ export function EmblaCarousel() {
 <TabsItem tab={TABS_LIBRARY.TABS.SOLID}>
 
 ```jsx highlight={13}
-import { createEffect } from 'solid-js'
+import { onMount } from 'solid-js'
 import createEmblaCarousel from 'embla-carousel-solid'
 
 export function EmblaCarousel() {
@@ -97,7 +97,7 @@ export function EmblaCarousel() {
     console.log(emblaApi.slidesInView())
   }
 
-  createEffect(() => {
+  onMount(() => {
     const api = emblaApi()
     if (api) api.on('slidesInView', logSlidesInView)
   })
@@ -180,7 +180,7 @@ export function EmblaCarousel() {
 
 ```html highlight={9}
 <script setup>
-  import { watchEffect } from 'vue'
+  import { onMounted } from 'vue'
   import emblaCarouselVue from 'embla-carousel-vue'
 
   const [emblaRef, emblaApi] = emblaCarouselVue()
@@ -190,7 +190,7 @@ export function EmblaCarousel() {
     emblaApi.off('slidesInView', logSlidesInViewOnce)
   }
 
-  watchEffect(() => {
+  onMounted(() => {
     if (emblaApi.value) emblaApi.value.on('slidesInView', logSlidesInViewOnce)
   })
 
@@ -202,7 +202,7 @@ export function EmblaCarousel() {
 <TabsItem tab={TABS_LIBRARY.TABS.SOLID}>
 
 ```jsx highlight={9}
-import { createEffect } from 'solid-js'
+import { onMount } from 'solid-js'
 import createEmblaCarousel from 'embla-carousel-solid'
 
 export function EmblaCarousel() {
@@ -213,7 +213,7 @@ export function EmblaCarousel() {
     emblaApi.off('slidesInView', logSlidesInViewOnce)
   }
 
-  createEffect(() => {
+  onMount(() => {
     const api = emblaApi()
     if (api) api.on('slidesInView', logSlidesInViewOnce)
   })
@@ -313,7 +313,7 @@ export function EmblaCarousel() {
 
 ```html highlight={3,10}
 <script setup>
-  import { watchEffect } from 'vue'
+  import { onMounted } from 'vue'
   import { EmblaCarouselType, EmblaEventType } from 'embla-carousel'
   import emblaCarouselVue from 'embla-carousel-vue'
 
@@ -326,7 +326,7 @@ export function EmblaCarousel() {
     console.log(`Embla just triggered ${eventName}!`)
   }
 
-  watchEffect(() => {
+  onMounted(() => {
     if (emblaApi.value) emblaApi.value.on('slidesInView', logEmblaEvent)
   })
 
@@ -346,7 +346,7 @@ export function EmblaCarousel() {
 <TabsItem tab={TABS_LIBRARY.TABS.SOLID}>
 
 ```jsx highlight={2,10}
-import { createEffect } from 'solid-js'
+import { onMount } from 'solid-js'
 import { EmblaCarouselType, EmblaEventType } from 'embla-carousel'
 import createEmblaCarousel from 'embla-carousel-solid'
 
@@ -360,7 +360,7 @@ export function EmblaCarousel() {
     console.log(`Embla just triggered ${eventName}!`)
   }
 
-  createEffect(() => {
+  onMount(() => {
     const api = emblaApi()
     if (api) api.on('slidesInView', logEmblaEvent)
   })
--- a/packages/embla-carousel-docs/src/content/pages/api/methods.mdx
+++ b/packages/embla-carousel-docs/src/content/pages/api/methods.mdx
@@ -58,12 +58,12 @@ export function EmblaCarousel() {
 
 ```html highlight={8}
 <script setup>
-  import { watchEffect } from 'vue'
+  import { onMounted } from 'vue'
   import emblaCarouselVue from 'embla-carousel-vue'
 
   const [emblaRef, emblaApi] = emblaCarouselVue()
 
-  watchEffect(() => {
+  onMounted(() => {
     if (emblaApi.value) console.log(emblaApi.value.slideNodes())
   })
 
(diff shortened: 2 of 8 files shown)


meirroth:
@davidjerleke Should we update the Vue example (and others) to use a ref, to demonstrate reactivity?
<TabsItem tab={TABS_LIBRARY.TABS.VUE}>
```html highlight={4}
<script setup>
  import emblaCarouselVue from 'embla-carousel-vue'

  const options = ref({ loop: true })

  const [emblaRef] = emblaCarouselVue(options)

  // ...
</script>

**Comment by meirroth** [comment]
@davidjerleke Reading the docs again, and looking at the source code I think we should update the the [Vue guide](https://www.embla-carousel.com/get-started/vue/) to mention:
1. [Options and plugin reactivity feature](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L46-L60). Perhaps with an example showing passing options as a ref.
2. Auto destroy [before unmount](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L42-L44).
3. Update the [API example](https://www.embla-carousel.com/get-started/vue/#accessing-the-carousel-api) to use onMounted instead of watchEffect.

Let me know what you think.

BTW, is it necessary to run .off on subscribed events before unmount?

**Comment by davidjerleke** [comment]
@meirroth thanks for your suggestions. 

> 1. [Options and plugin reactivity feature](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L46-L60). Perhaps with an example showing passing options as a ref.

Maybe adding a heading below [Global options](https://embla-carousel.com/api/options/#global-options) and name it **Reactive options**? That section could demonstrate a code example off how to work with reactive options for all reactive Embla wrappers: `React`, `Vue`, `Solid` and `Svelte`.

> 2. Auto destroy [before unmount](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L42-L44).

I agree. We should add this information on all get started pages `get-started/<embla-package>`. Maybe it makes sense to add a heading after [Adding plugins](https://www.embla-carousel.com/get-started/vue/#adding-plugins) called **Destroying a carousel**?

> 3. Update the [API example](https://www.embla-carousel.com/get-started/vue/#accessing-the-carousel-api) to use onMounted instead of watchEffect.

Yes, I agree. Unfortunately the watchEffect examples are sprinkled out so we need to find these places:

**Events**
- [Adding event listeners](https://www.embla-carousel.com/api/events/#adding-event-listeners)
- [Removing event listeners](https://www.embla-carousel.com/api/events/#removing-event-listeners)
- [TypeScript](https://www.embla-carousel.com/api/events/#typescript)

**Methods**
- [Calling methods](https://www.embla-carousel.com/api/methods/#calling-methods)
- [TypeScript](https://www.embla-carousel.com/api/methods/#typescript)

**Plugins**
- [Calling methods](https://www.embla-carousel.com/api/plugins/#calling-methods)
- [Adding event listeners](https://www.embla-carousel.com/api/plugins/#adding-event-listeners)

> BTW, is it necessary to run `.off` on subscribed events before unmount or does `destroy` handle that?

`destroy` doesn't handle that but I think it should. To achieve this it has ti be added after [this line](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EmblaCarousel.ts#L154). We also need to add a `clear` method [here](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EventHandler.ts#L55) like so:

```tsx
function clear(): void {
  listeners = {}
}
```

Best,
David

meirroth:
@davidjerleke I'll work on all of this once I get the repo running localy.


> `destroy` doesn't handle that but I think it should. To achieve this it has to be added after [this line](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EmblaCarousel.ts#L154). We also need to add a `clear` method [here](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EventHandler.ts#L55) like so:
> ```
> function clear(): void {
>   listeners = {}
> }
> ```
> 

Can you please clarify what you meant by "**it has to be added after [this line](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EmblaCarousel.ts#L154).**" 

1. What exactly should be added after that line? 
2. Are we updating the core package or only in the vue package?
3. Wouldn't clearing event listeners be handled within the destroy method?
4. I'm guessing we'll handle removing event listeners by calling .off on all subscribed listeners, or would you do it differently?

meirroth:
Also, trying to understand if `engine.eventStore.clear()` clears event listeners, if so, it looks like it's already handled by `deActivate` which `destory` calls.

meirroth:
> > 1. [Options and plugin reactivity feature](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L46-L60). Perhaps with an example showing passing options as a ref.
> 
> Maybe adding a heading below [Global options](https://embla-carousel.com/api/options/#global-options) and name it **Reactive options**? That section could demonstrate a code example off how to work with reactive options for all reactive Embla wrappers: `React`, `Vue`, `Solid` and `Svelte`.
> 
> > 2. Auto destroy [before unmount](https://github.com/davidjerleke/embla-carousel/blob/09da7a68fd7629958f1271e5b3c9c04acbce704b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts#L42-L44).
> 
> I agree. We should add this information on all get started pages `get-started/<embla-package>`. Maybe it makes sense to add a heading after [Adding plugins](https://www.embla-carousel.com/get-started/vue/#adding-plugins) called **Destroying a carousel**?
> 
> > 3. Update the [API example](https://www.embla-carousel.com/get-started/vue/#accessing-the-carousel-api) to use onMounted instead of watchEffect.
> 
> Yes, I agree. Unfortunately the watchEffect examples are sprinkled out so we need to find these places ...

@davidjerleke Done 1 and 3.

For 2, I'm just thinking whether instead of adding a new section like you suggested for mentioning that the framework wrapper auto destroys the carousel we should add an opening line at the top of the page, above "Start by installing the Embla Carousel...", something like this:

> # Vue
> 
> Embla Carousel provides a wrapper for Vue that ensures seamless integration of the carousel into your Vue project and automatic cleanup on component unmount.
> 
> Start by installing the Embla Carousel **npm package** and add it to your dependencies.
> 
Thoughts?

For 3, I'm only familiar enough to provide  examples for Vanilla and Vue, can you please help me with the other frameworks?

davidjerleke:
> Can you please clarify what you meant by "**it has to be added after [this line](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel/src/components/EmblaCarousel.ts#L154).**"
> 
> 1. What exactly should be added after that line?

I can look into this as it's a bit hard to exmplain.

> 2. Are we updating the core package or only in the vue package?

The core package 🙂. 

> 3. Wouldn't clearing event listeners be handled within the destroy method?

Yes they should.

> 4. I'm guessing we'll handle removing event listeners by calling .off on all subscribed listeners, or would you do it differently?

We can under the lifecycle of a carousel add (.on) and remove (.off) listeners but they should be cleared when the carousel is `destroyed`. This isn't a big problem because I don't get bug reports about it but still we should do the housekeeping.

meirroth:
@davidjerleke Thank you for your guidance! I've update the title, and added an introduction line to all wrappers.

What's left for this PR is the reactive examples.

<hr>

We can continue to discuss removing event listeners here for continuity sake, but I think I'll open a new PR if any changes needed.

Surfacing what I wrote earlier: does `mediaHandlers.clear()` or `deActivate() --> engine.eventStore.clear()` called by the [`destory()` method](https://github.com/davidjerleke/embla-carousel/blob/baf030b0bf7ab82976576c7346e71a92485cf062/packages/embla-carousel/src/components/EmblaCarousel.ts#L152-L153) handle clearing event listeners? If so, I think we're good as is, and there's no need to manually run `.off` on subscribed events before unmount, [as I initially asked](https://github.com/davidjerleke/embla-carousel/pull/836#issuecomment-2073201078).

davidjerleke:
@meirroth fantastic work here. I added the missing examples. Now we only have to update the [api/plugins](https://www.embla-carousel.com/api/plugins/) page because plugins are reactive too 😅. But let's do that in a separate PR. 

About this:

> Surfacing what I wrote earlier: does `mediaHandlers.clear()` or `deActivate() --> engine.eventStore.clear()` called by the [`destory()` method](https://github.com/davidjerleke/embla-carousel/blob/baf030b0bf7ab82976576c7346e71a92485cf062/packages/embla-carousel/src/components/EmblaCarousel.ts#L152-L153) handle clearing event listeners? If so, I think we're good as is, and there's no need to manually run `.off` on subscribed events before unmount, [as I initially asked](https://github.com/davidjerleke/embla-carousel/pull/836#issuecomment-2073201078).

I will create a PR to show you what I mean. Stay tuned 👍.

meirroth:
@davidjerleke Awesome, thank you!

meirroth:
> > Surfacing what I wrote earlier: does `mediaHandlers.clear()` or `deActivate() --> engine.eventStore.clear()` called by the [`destory()` method](https://github.com/davidjerleke/embla-carousel/blob/baf030b0bf7ab82976576c7346e71a92485cf062/packages/embla-carousel/src/components/EmblaCarousel.ts#L152-L153) handle clearing event listeners? If so, I think we're good as is, and there's no need to manually run `.off` on subscribed events before unmount, [as I initially asked](https://github.com/davidjerleke/embla-carousel/pull/836#issuecomment-2073201078).
> 
> I will create a PR to show you what I mean. Stay tuned 👍.

@davidjerleke Was this addressed here? 😏 https://github.com/davidjerleke/embla-carousel/pull/871/files#diff-cb6e53e6d149a1a6784922356eaff9e7a5f0a43a0d30203fec752c39b60aeaf0R155

davidjerleke:
@meirroth yes! And [here](https://github.com/davidjerleke/embla-carousel/pull/871/files#diff-065ad6a5c9a4b1ef24b4d5de47f8a9c043f1438fd9dcdcf681ca8f7b78af4cffR58-R60) of course.

meirroth:
@davidjerleke Great work! Thank you

Issue #39, opened 2019-12-23
[Feat]: Add a fade plugin


dmorda:
Love the carousel and was able to get it working right away. The one thing that I need that it doesn't provide is a way to have a scroll option for single/multiple slides and a fade option to handle next/previous transition when it's just showing a single slide at a time.

Is that something that has been requested before?

davidcetinkaya:
Hi Damon (@dmorda),
Merry Christmas, and thank you for opening this issue 🎅. 

About this:

> A way to have a scroll option for single/multiple slides

Would you mind explaining what you mean a bit more in detail? I need more information in order to answer this question, but I'm going to take a guess and ask if you've checked out the [`slidesToScroll`](https://github.com/davidcetinkaya/embla-carousel#slidestoscroll) option? Here's a [CodeSandbox](https://codesandbox.io/s/embla-carousel-slidestoscroll-1g4rk) that demonstrates how you can achieve this.

The answer to your question about the fade option is no, it has not been requested yet. It's not on the roadmap of Embla but maybe you can achieve this by hooking onto the classname toggling for selected slides? Please have a look at the documentation [here](https://github.com/davidcetinkaya/embla-carousel#selectedclass--type-string).

I hope any of this helps!

Best,
David

dmorda:
Thanks for the quick response and a Merry Christmas to you as well. What I'm trying to achieve is a Fade effect , as an example view [this page](https://kenwheeler.github.io/slick/) and search for the word "Fade". That will provide an example of what I'm trying to accomplish.

Let me know if that helps!

davidcetinkaya:
Hi again Damon (@dmorda),
Thank you for the clarification, that helps 👍.

The vision for Embla is physics simulated carousels that move in a fluid and natural manner. The assumption here is that the carousel should be scrollable, and this doesn't play well with fade sliders that don't feel natural to drag/swipe. This is why a fade option is not a part of the Embla vision.

But with that said, **I've setup a [CodeSandbox example](https://codesandbox.io/s/embla-carousel-github-issue-39-sdkul) for you**, where I've hacked the fade option by tweaking some CSS.

I hope that you'll find it useful!

Best,
David

dmorda:
That small tweak now lets me re-use Embla for both the fade version of a carousel and the normal sliding version. Thanks so much, happy holidays!

davidcetinkaya:
Hi Damon (@dmorda),

[...]

PR #871, opened 2024-05-13
[Feat]: Add a fade plugin

davidjerleke:
- Closes: #39.


freddieerg:
Hey @davidjerleke, thanks for making this plugin! One thing I would like to see is an option to change how the slide fade happens. I have a use case where I would like to keep the outgoing slide at 100% opacity the incoming slide has reached 100% opacity, this way I can avoid the flash of background when both slides are at that 50% opacity point. I'd be happy to make a PR for this, but do you have any idea how you'd like to see this exposed as a user option? Thanks :)

Issue #837, opened 2024-04-23
[Bug]: Jest testing bug return TypeError: undefined is not a function


KareyM:
### Which variants of Embla Carousel are you using?

- [X] embla-carousel **(Core)**
- [ ] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

When writing test for the `EmblaCarousel`, i'm getting this error. Not sure what it is that's causing the issue. However, it seem to work just fine within the browser

```ts
TypeError: undefined is not a function
        at Array.map (<anonymous>)

      110 |   const DELAY = 4000;
      111 |   const plugins = [EmblaCarouselAutoplay({ delay: DELAY })];
    > 112 |   const emblaApi = EmblaCarousel(emblaNode, options, plugins);
          |                                 ^
      113 |   return emblaApi;
      114 | };
      115 |
```

### Expected Behavior

During a test, calling `EmblaCarousel` should behave the same work it does during the browser

### Additional Context

I'm using the TS version of this and downloaded it through `yarn`

### Version

v8.0.2

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

davidjerleke:
@KareyM this is not a bug. You need to mock the following API:s to solve the problem:

- `IntersectionObserver`
- `matchMedia`
- `resizeObserver`

Feel free to [grab the mocks](https://github.com/davidjerleke/embla-carousel/tree/master/packages/embla-carousel/src/__tests__/mocks) that the core library `embla-carousel` is using.

Best,
David

KareyM:
Thank you!!

PR #876, opened 2024-05-17
Add better error message for missing mocks

JuhG:
Right now it's hard to figure out what's wrong here:
Screenshot 2024-05-17 at 13 01 09
"undefined is not a function"

Almost everyone will run into this after upgrading to the new major version. Maybe we could add a better error message.
Perhaps linking to [this comment](https://github.com/davidjerleke/embla-carousel/issues/837#issuecomment-2073969586) or directly to the mocks.

Looking at caniuse stats, it's pretty unrealistic that people will run into these errors unless running tests.
What do you think?

### DIFF
--- a/packages/embla-carousel/src/components/OptionsHandler.ts
+++ b/packages/embla-carousel/src/components/OptionsHandler.ts
@@ -31,6 +31,9 @@ export function OptionsHandler(ownerWindow: WindowType): OptionsHandlerType {
   }
 
   function optionsMediaQueries(optionsList: OptionsType[]): MediaQueryList[] {
+    if (!('matchMedia' in ownerWindow)) {
+      throw new Error('matchMedia does not exist on window, mock it if running tests')
+    }
     return optionsList
       .map((options) => objectKeys(options.breakpoints || {}))
       .reduce((acc, mediaQueries) => acc.concat(mediaQueries), [])


davidjerleke:
@JuhG thank you for your efforts and sorry about the delayed response. While I think this is a good idea, I would like to see a different approach to solve this. Because the current solution will make it to the production bundle increasing the bundle size. I've been planning to create separate bundles for dev and prod where the dev bundle can include helpful error messages like this and many more. But I haven't had the time to change this yet.

JuhG:
@davidjerleke that's true
maybe you could try to add something like https://github.com/alexreardon/tiny-warning or just manually checking NODE_ENV:
```tsx
if ('production' !== process.env.NODE_ENV) {
  warning(condition, 'My cool message that takes up a lot of kbs');
}
```
these messages disappear when bundlers do their thing, so realistically they won't increase the bundle size (so no need to create separate dev and prod bundles)

davidjerleke:
@JuhG are you sure these if statements disappear automatically? 

There are different bundlers out there like Webpack, Esbuild, RollupJS etc. Do all of them really strip these if statements automatically without the need of using plugins and configuring them? I would like to have proof that this is the case before I proceed with this.

davidjerleke:
@JuhG as mentioned I appreciate the effort but would need hard evidence that the most common build tools strip this automatically.

Issue #910, opened 2024-06-11
[Bug]: Auto-height not detecting dynamic element changes


lucaspbordignon:
### Which variants of Embla Carousel are you using?

- [ ] embla-carousel **(Core)**
- [ ] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [X] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-fade
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

The bug occurs when using the `embla-carousel-auto-height` package and the carousel elements have a dynamically set height

As an example, by using an `<img />` tag with unknown dimensions upfront

What happens is that the carousel uses the initial height as the slide's height, prior to the element being loaded

### Expected Behavior

The expected behavior is for `embla-carousel` to automatically calculate the element's height when navigating to it

### Additional Context

Additionally, when resizing the window **after** the elements have their final height defined, we see the carousel being re-initialized, which ends up solving the problem

### What browsers are you seeing the problem on?

All browsers

### Version

v8.1.3

### CodeSandbox

https://codesandbox.io/p/sandbox/embla-carousel-auto-height-react-lazy-images-p6f2f9

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

lucaspbordignon:
@davidjerleke - Additionally, I've added #911 to the project, intending to solve the issue mentioned. Let me know if you think we should go to any other direction there

davidjerleke:
Hi @lucaspbordignon,

The default behaviour for Embla is that it will only react to slide size changes that affect the chosen scroll axis. This means that a horizontal carousel only re-initialises when the container and/or slide widths change. In contrast, a vertical carousel only re-initialises when the container and/or slide heights change.

However, you can customize the resize behavior easily with the [watchResize](https://www.embla-carousel.com/api/options/#watchresize) option. The most simple way to solve this would be to re-initialise the carousel when any dimension change, whether it’s heights or widths:

You should be able to solve your it like this:

```tsx
const emblaApi = EmblaCarousel({
  watchResize: () => {
[...]

PR #911, opened 2024-06-11
embla-carousel-auto-height | Start using ResizeObserver for height calculations

lucaspbordignon:
## Context

After an initial investigation, have noticed that the `slideHeights` element for `embla-carousel-auto-height` was being calculated only when initializing the plugin, which does not detect dynamic changes to the carousel elements

The proposed solution leverages [ResizeObsevers](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe) to detect changes to the slide heights and re-calculate them properly

Solving https://github.com/davidjerleke/embla-carousel/issues/910

## Additional Details

ResizeObservers do have relevant coverage with all [major browsers](https://caniuse.com/resizeobserver), being a good candidate for the solution. Open for new ideas as always on how we could increase support


davidjerleke:
Will be replaced by [this solution](https://github.com/davidjerleke/embla-carousel/issues/910#issuecomment-2180511909).

PR #929, opened 2024-07-03
refactor(vue): optimize reactivity handling and reduce bundle size

negezor:
Hello, I accidentally came across your module and, after looking at the source code, I really liked it. I would like to thank you for it.

While examining the Vue implementation, I noticed a few areas for improvement. Here is the list:
1. Use `shallowRef` instead of `ref`. It is standard practice to use it for `HTMLElement` and heavy objects that do not require reactivity. We just need to observe the value setting in this case.
2. Use simple variables to store `storedOptions` and `storedPlugins` that do not need reactivity. After all, we only add `watch` for `Ref` objects.
3. The first two improvements also slightly reduce the minified bundle size, as we only call `isRef` once and remove unnecessary accesses to `.value`.

### DIFF
--- a/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
+++ b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
@@ -1,4 +1,4 @@
-import { Ref, ref, isRef, watch, onMounted, onBeforeUnmount } from 'vue'
+import { Ref, isRef, watch, onMounted, onBeforeUnmount, shallowRef } from 'vue'
 import {
   areOptionsEqual,
   arePluginsEqual,
@@ -19,10 +19,10 @@ function emblaCarouselVue(
   options: EmblaOptionsType | Ref<EmblaOptionsType> = {},
   plugins: EmblaPluginType[] | Ref<EmblaPluginType[]> = []
 ): EmblaCarouselVueType {
-  const storedOptions = ref(isRef(options) ? options.value : options)
-  const storedPlugins = ref(isRef(plugins) ? plugins.value : plugins)
-  const emblaNode = ref<HTMLElement>()
-  const emblaApi = ref<EmblaCarouselType>()
+  const storedOptions = shallowRef(isRef(options) ? options.value : options)
+  const storedPlugins = shallowRef(isRef(plugins) ? plugins.value : plugins)
+  const emblaNode = shallowRef<HTMLElement>()
+  const emblaApi = shallowRef<EmblaCarouselType>()
 
   function reInit() {
     if (!emblaApi.value) return
--- a/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
+++ b/packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
@@ -1,4 +1,12 @@
-import { Ref, isRef, watch, onMounted, onBeforeUnmount, shallowRef } from 'vue'
+import {
+  Ref,
+  MaybeRef,
+  isRef,
+  watch,
+  onMounted,
+  onBeforeUnmount,
+  shallowRef
+} from 'vue'
 import {
   areOptionsEqual,
   arePluginsEqual,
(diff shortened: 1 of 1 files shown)


davidjerleke:
Thanks for your contribution @negezor.

> 1. Use `shallowRef` instead of `ref`. It is standard practice to use it for `HTMLElement` and heavy objects that do not require reactivity. We just need to observe the value setting in this case.

Thanks. Makes sense 👍. 

> 2. Use simple variables to store `storedOptions` and `storedPlugins` that do not need reactivity. After all, we only add `watch` for `Ref` objects.

I'm not following here. Options and plugins **should be reactive**. Whenever options or plugins change, the carousel should re-initialize. Or am I misunderstanding what you mean by this?

> 3. The first two improvements also slightly reduce the minified bundle size, as we only call `isRef` once and remove unnecessary accesses to `.value`.

Thanks!

negezor:
> I'm not following here. Options and plugins should be reactive. Whenever options or plugins change, the carousel should re-initialize. Or am I misunderstanding what you mean by this?

Options and plugins will be reactive. We don't just need to store them in a reactive variable, because we only listen for changes to the passed arguments.

davidjerleke:
> Options and plugins will be reactive. We don't just need to store them in a reactive variable, because we only listen for changes to the passed arguments.

@negezor thanks for clarifying. I think this is a nice refactor of the Vue wrapper so great work 👍🏻. Have you tested the wrapper with your changes?

@meirroth would you like to skim the code and add anything?

negezor:
> @negezor thanks for clarifying. I think this is a nice refactor of the Vue wrapper so great work 👍🏻. Have you tested the wrapper with your changes?

@davidjerleke yes. I used the fork of version 8.1.5 with these changes on my stage, as well as later on in production.

packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
```diff
@@ -19,42 +19,46 @@ function emblaCarouselVue(
...
   plugins: EmblaPluginType[] | Ref<EmblaPluginType[]> = []
 ): EmblaCarouselVueType {
-  const storedOptions = ref(isRef(options) ? options.value : options)
-  const storedPlugins = ref(isRef(plugins) ? plugins.value : plugins)
-  const emblaNode = ref<HTMLElement>()
-  const emblaApi = ref<EmblaCarouselType>()
+  const isRefOptions = isRef(options)
+  const isRefPlugins = isRef(plugins)
+
+  let storedOptions = isRefOptions ? options.value : options
```
sadeghbarati, line 25: I would use `unref` instead 

> This is a sugar function for `val = isRef(val) ? val.value : val.`

sadeghbarati:
Using `shallowRef` is LGTM 👌

meirroth:
Looks good, thanks 👍

davidjerleke:
Just tag me here when this is approved @meirroth and @sadeghbarati. Thanks 🙏🏻.

packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
negezor: Changed to `MaybeRef`, good catch. We are already checking for `isRef`, so we might not need `unref`, as it would result in a redundant check. Although this is a micro-optimization, why not?

sadeghbarati:
```ts
const 
```

sadeghbarati: #### Now I understand `isRef`, you add watch for `options` and `plugins` changes with `isRef` condition

But how about the `watchEffect`?

sadeghbarati (approved):

sadeghbarati:
@davidjerleke done 😄

meirroth (approved):
Looks good 👍 Tested in my project and works fine.

packages/embla-carousel-vue/src/components/emblaCarouselVue.ts
negezor: It doesn’t make sense to use `watchEffect` in this case, since we only need to track specific changes to `plugins` or `options`. Otherwise, these checks will be called for any reactive change, which will cause more performance loss.

davidjerleke:
Will be included in the `v8.1.7` release.

davidjerleke:
@negezor, @sadeghbarati, @meirroth these improvements were just included in the [v8.1.7](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.1.7) release.

PR #947, opened 2024-07-20
fix(react): export EmblaViewportRefType from bundle

junlarsen:
Hi David, thanks for your awesome work on Embla!

I recently wrote a small wrapper component around the react plugin and found it would be beneficial to pass around the Embla viewport ref type (for example uses like forwardRef).

This small patch re-exports EmblaViewportRefType from embla-carousel-react. Let me know if there are any questions!

### DIFF
--- a/packages/embla-carousel-react/src/components/useEmblaCarousel.ts
+++ b/packages/embla-carousel-react/src/components/useEmblaCarousel.ts
@@ -10,7 +10,7 @@ import EmblaCarousel, {
   EmblaPluginType
 } from 'embla-carousel'
 
-type EmblaViewportRefType = <ViewportElement extends HTMLElement>(
+export type EmblaViewportRefType = <ViewportElement extends HTMLElement>(
   instance: ViewportElement | null
 ) => void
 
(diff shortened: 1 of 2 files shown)


davidjerleke (changes requested):
Hi @junlarsen,

Thanks for your contribution! I think these changes make sense. We also need to export it from the `index` file:

https://github.com/davidjerleke/embla-carousel/blob/9d1cc8a235cbbcac783e270d44181d83076ed938/packages/embla-carousel-react/src/index.ts#L1

The reason why is because package managers like `pnpm` have introduced new behavior to the frontend world. For example, `pnpm` doesn’t allow imports from any other file than the `index` file, so any imports with paths to specific files will fail.

junlarsen:
Right, missed that one. I'll get on it tonight :)

davidjerleke (approved):

davidjerleke:
@junlarsen this is approved. Nicely done 👍🏻. Feel free to squash the commits into a single commit. Otherwise I will do it before merging this PR.

davidjerleke:
Thanks @junlarsen 👍🏻!

davidjerleke:
@junlarsen your fix has been included in [`v8.1.8`](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.1.8) which I just released.

junlarsen:
Great, thank you!

Issue #909, opened 2024-06-11
[Bug]: Add nodenext compatibility for ESM imports


fethca:
### Which variants of Embla Carousel are you using?

- [x] embla-carousel **(Core)**
- [X] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-fade
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

The current version of Embla Carousel lacks compatibility with the `nodenext` module resolution setting in TypeScript. It requires explicit file extensions in ESM imports. This will cause the build to fail with the error : 

```
node_modules/.pnpm/embla-carousel@8.1.4/node_modules/embla-carousel/esm/components/EmblaCarousel.d.ts:1:28 - error TS2834: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

import { EngineType } from './Engine'
```

The tsconfig I use to test it : 

```
{
  "compilerOptions": {
    "outDir": "dist",
    "module": "nodenext",
    "target": "ESNext",
    "strict": true,
    "resolveJsonModule": true,
    "jsx": "react-jsx",
  }
}
```

### Expected Behavior

No error on build

### Additional Context

I did a PoC of what would need to be done on this [fork](https://github.com/fethca/embla-carousel) and released [this version](https://www.npmjs.com/package/@fethcat/embla-carousel) on npm to test it on my applications. From what I saw most of the job is to explicitely add proper file extension in import path so `import { AnimationsType } from './Animations'` becomes `import { AnimationsType } from './Animations.js'`. This change will break the rollup config for writing types cause it's used to add `.ts` to paths, I just changed it to `fs.writeFileSync(esmTypesFilePath, esmTypesFile)`. 

I only adapted `embla-carousel` & `embla-carousel-react` packages as those are the ones I needed for my app. 

### Version

8.1.4

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

davidjerleke:
[...]

Issue #780, opened 2024-02-29
[Bug]: Cannot find module 'embla-carousel/components/Options' or its corresponding type declarations. TypeScript, React with Vite


rxtsel:
### Which variants of Embla Carousel are you using?

- [X] embla-carousel **(Core)**
- [X] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

It seems that the issue stems from the way `OptionsType` is imported in the` useEmblaCarousel.d.ts` file within the `embla-carousel-react` folder in `node_modules`. Instead of importing directly from `'embla-carousel/components/Options'`, it should be imported directly from the `EmblaOptionsType` type.

### Expected Behavior

![expect](https://github.com/davidjerleke/embla-carousel/assets/85462420/2043f427-099b-4e6c-96f0-7e363336ead0)

### Received behavior

![return](https://github.com/davidjerleke/embla-carousel/assets/85462420/862b1672-1694-4dc0-a01d-fe1d442a8e92)

### Additional Context

The bug occurs when I build:

![image](https://github.com/davidjerleke/embla-carousel/assets/85462420/381b7504-3d55-4646-b1a2-b3b9743ef7f4)

### Version

v8.0.0

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

davidjerleke:
Hi @rxtsel,

Thanks for the bug report. It seems like many devs are having this problem. 

- See this: https://github.com/davidjerleke/embla-carousel/discussions/726#discussioncomment-8331396

Setting `moduleResolution` to `node` in `tsconfig` seems to solve it, but I don’t know why yet.

> It seems that the issue stems from the way `OptionsType` is imported in the` useEmblaCarousel.d.ts` file within the `embla-carousel-react` folder in `node_modules`. Instead of importing directly from `'embla-carousel/components/Options'`, it should be imported directly from the `EmblaOptionsType` type.

Have you tried your solution? 
Which package manager are you using?

- Could be related to #778 

Best,
David

[...]

Issue #778, opened 2024-02-29
[Feat]: Add TypeScript moduleResolution: bundler support


crashton28:
### Which variants of Embla Carousel are you using?

- [x] embla-carousel **(Core)**
- [x] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [x] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

In attempting to use the `isPlaying` method as well as the `autoplay:stop` and `autoplay:play` event listeners the types are not found.  This project is currently within a monorepo and using `pnpm`.

We are able to resolve the issue by updating our `tsconfig` with the following, however we are not able to make this change as we need `moduleResolution` to be set to `bundler` for our builds.

```
"moduleResolution": "node10",
"module": "ESNext"
```

Screenshots of the errors within VSCode

![Screenshot 2024-02-29 at 9 58 20 AM](https://github.com/davidjerleke/embla-carousel/assets/15710355/b8cab703-330d-426e-96e3-f5c731a804d8)

![Screenshot 2024-02-29 at 9 58 39 AM](https://github.com/davidjerleke/embla-carousel/assets/15710355/413620b1-96f6-4c3e-9906-e347cadcea6b)

![Screenshot 2024-02-29 at 9 58 48 AM](https://github.com/davidjerleke/embla-carousel/assets/15710355/d7577f25-33c4-4172-ae71-e06e3204110b)

### Expected Behavior

Types would have been found

### Additional Context

I have also tried...

### Version

v8.0.0

### CodeSandbox

[CodeSandbox: Monorepo with TS Error](https://codesandbox.io/p/devbox/embla-carousel-autoplay-react-forked-7tkpmg?file=%2Fpackages%2Fui%2Fsrc%2Fjs%2FEmblaCarousel.tsx%3A62%2C48)

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

davidjerleke:
@crashton28 thanks for your bug report. Please create a minimal reproduction environment as it’s mandatory if you want me to look into this.

Best,
David

[...]

PR #1014, opened 2024-09-23
[react] re-export `EmblaCarouselType`

rawnly:
Right now if the developer wants to use `EmblaCarouselType` must install `embla-carousel` too. To avoid this I'm just re-exporting it in the barrel file.

### Usage
```diff
-import type { UseEmblaCarouselType } from "embla-carousel-react";
+import type { EmblaCarouselType } from "embla-carousel-react";
-type EmblaCarouselType = NonNullable<UseEmblaCarouselType[1]>;

const options: EmblaCarouselType = {}
```

### DIFF
--- a/packages/embla-carousel-react/src/index.ts
+++ b/packages/embla-carousel-react/src/index.ts
@@ -2,4 +2,7 @@ export {
   UseEmblaCarouselType,
   EmblaViewportRefType
 } from './components/useEmblaCarousel'
+
 export { default } from './components/useEmblaCarousel'
+
+export { EmblaCarouselType } from 'embla-carousel'


davidjerleke:
Hi @rawnly,

As much as I appreciate the effort, I would recommend you to always start a discussion before creating PR's. This is for your own sake so you don't spend valuable time doing stuff that might not be on the roadmap or similar.

The `EmblaCarouselType` and other types were exported in earlier versions for convenience but they were later removed. Let me tell you that it's really really messy to support all different combinations of package managers and TypeScript configs (`moduleResolution: 'node' | 'bundler'` etc). But you can mainly thank the `pnpm` package manager for this: They introduced a new package manager behavior where nested dependencies aren't accessible like they are when you use `npm/yarn` so re-exports simply don't work. This was also one of the reasons for the huge type mess:

- #909
- #780
- #778
- #726
- and more...

Now we finally got a working setup for all possible combinations and I'm not going to change that, risking me getting stuck in a loop of endless bug reports/questions about types. Devs using `pnpm` get confused when the exports are there in the code but don't work and I simply don't have time answering the same questions over and over again.

You simply have to see this as installing a type package like `@types/react`, because that's kind of what you're doing when installing `embla-carousel` in order to use its types. This is because all official Embla packages have first class TypeScript support, meaning they're written in TypeScript.

I hope you understand.

Best,
David

rawnly:
Thank you for such a complete response! Don't worry about the effort. To be honest it was a super fast change I did using the github editor without even cloning the repo 😅 

I appreciate your explanation!

PR #1041, opened 2024-10-22
[Feat]: Add react v19 to peer deps

Yonom:
Thanks a lot for this package

I get the following warning from pnpm after upgrading to NextJS v15 (released today), which installs react 19 RC

React 19 is technically still in RC stage, but since a stable version of a major Metaframework is installing it, it's a sign that it's ready for production use and will now be used by many projects.

```ts
└─┬ embla-carousel-react 8.3.0
  └── ✕ unmet peer react@"^16.8.0 || ^17.0.1 || ^18.0.0": found 19.0.0-rc-65a56d0e-20241020
```

### DIFF
--- a/packages/embla-carousel-react/package.json
+++ b/packages/embla-carousel-react/package.json
@@ -64,7 +64,7 @@
     "embla-carousel-reactive-utils": "8.3.0"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.1 || ^18.0.0"
+    "react": "^16.8.0 || ^17.0.1 || ^18.0.0 || ^19.0.0 || ^19.0.0-rc"
   },
   "exports": {
     "./package.json": "./package.json",


davidjerleke:
Thank you for your contribution @Yonom, much appreciated 👍.

### I have one question
I haven't checked our `React 19` so I don't know if it has any breaking changes that will affect/break the [`useEmblaCarousel`](https://github.com/davidjerleke/embla-carousel/blob/master/packages/embla-carousel-react/src/components/useEmblaCarousel.ts) hook. Will it be compatible without any changes?

Yonom:
I couldn't spot any issues in my installation of it!

ricardo-a-alves-alb:
By looking at the official [upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument), the only thing that seems to "apply" to this codebase is the `useRef` part, which already is implemented this way on `useEmblaCarousel` hook.

davidjerleke:
@ricardo-a-alves-alb do you mean that the `useRef` code in `useEmblaCarousel` will work with React 19 as is or that it needs to change?

ricardo-a-alves-alb:
> @ricardo-a-alves-alb do you mean that the `useRef` code in `useEmblaCarousel` will work with React 19 as is or that it needs to change?

As-is, it should work properly. The difference from 18 to 19 is that `useRef()` does not allow no arguments passed. Since `useEmblaCarousel` always passes a default value to `useRef()`, it should work with no issues.

davidjerleke:
@ricardo-a-alves-alb thanks for digging into this. I will merge this soon. @Yonom could you check if this PR needs to be rebased and do that on the latest master in case it’s necessary?

davidjerleke:
Thanks @Yonom for your contribution and @ricardo-a-alves-alb for the reference 👍.

davidjerleke:
@Yonom and @ricardo-a-alves-alb I just released this in [v8.3.1](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.3.1).

ricardo-a-alves-alb:
> @Yonom and @ricardo-a-alves-alb I just released this in [v8.3.1](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.3.1).

Nice! One more dependency updated so I can upgrade to Next.js 15 later on 😄

PR #1054, opened 2024-11-06
[Feat]: Trigger event (autoplay:select) on autoplay index update

Mitch-At-Work:
Emit autoplay:triggered event on autoplay index change as discussed in:

- https://github.com/davidjerleke/embla-carousel/discussions/1048

Event is emitted after index change; listeners can retrieve the latest selectedScrollSnap from emblaApi regardless of direction to be compatible with upcoming autoplay direction feature.

### DIFF
--- a/packages/embla-carousel-autoplay/src/components/Autoplay.ts
+++ b/packages/embla-carousel-autoplay/src/components/Autoplay.ts
@@ -13,6 +13,7 @@ declare module 'embla-carousel' {
   interface EmblaEventListType {
     autoplayPlay: 'autoplay:play'
     autoplayStop: 'autoplay:stop'
+    autoplaySelect: 'autoplay:select'
   }
 }
 
@@ -165,6 +166,8 @@ function Autoplay(userOptions: AutoplayOptionsType = {}): AutoplayType {
     } else {
       emblaApi.scrollTo(0, jump)
     }
+
+    emblaApi.emit('autoplay:select')
   }
 
   const self: AutoplayType = {
--- a/packages/embla-carousel-docs/src/content/pages/plugins/autoplay.mdx
+++ b/packages/embla-carousel-docs/src/content/pages/plugins/autoplay.mdx
@@ -193,3 +193,11 @@ Once: <BrandPrimaryText>`no`</BrandPrimaryText>
 Fires when autoplay stops playing.
 
 ---
+
+### autoplay:select
+
+Once: <BrandPrimaryText>`no`</BrandPrimaryText>
+
+Fires directly after autoplay selects a new index.
+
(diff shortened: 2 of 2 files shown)


packages/embla-carousel-autoplay/src/components/Autoplay.ts
```diff
@@ -13,6 +13,7 @@ declare module 'embla-carousel' {
   interface EmblaEventListType {
     autoplayPlay: 'autoplay:play'
     autoplayStop: 'autoplay:stop'
+    autoplayTriggered: 'autoplay:triggered'
```
davidjerleke, line 16: What do you think about renaming this to `autolplay:select`? That would be consistent with the core API but is a clear distinction that it's triggered by `Autoplay`.

packages/embla-carousel-autoplay/src/components/Autoplay.ts
```diff
@@ -165,6 +166,7 @@ function Autoplay(userOptions: AutoplayOptionsType = {}): AutoplayType {
     } else {
       emblaApi.scrollTo(0, jump)
     }
+    emblaApi.emit('autoplay:triggered')
```
davidjerleke, line 169: Nitpick: Maybe add a space above this line?

davidjerleke (changes requested):
Nicely done.

Just two minor comments. Additionally, we should add the event [here](https://www.embla-carousel.com/plugins/autoplay/#events) in the docs.

packages/embla-carousel-autoplay/src/components/Autoplay.ts
Mitch-At-Work: Sounds good! I'll make these updates today

packages/embla-carousel-autoplay/src/components/Autoplay.ts
```diff
@@ -13,6 +13,7 @@ declare module 'embla-carousel' {
   interface EmblaEventListType {
     autoplayPlay: 'autoplay:play'
     autoplayStop: 'autoplay:stop'
+    autoplayTriggered: 'autoplay:select'
```
Mitch-At-Work, line 16: Todo: Fix up type name.

davidjerleke (approved):

davidjerleke:
Thanks for your contribution @Mitch-At-Work. Feel free to squash the commits into to a single commit with the message:

`Implement #1048.`

If not, I will do it when possible.

Cheers 👍🏻!

davidjerleke:
@Mitch-At-Work great work and thanks for your contribution. This feature will be included in the next release 🥳.

PR #1062, opened 2024-11-18
[Feat]: Autoplay - Delay between transitions per slide + Possibility to create a progress bar

davidjerleke:
Closes: 
- #743
- #1059


davidjerleke:
Note to self: Add the updated delay option to the docs.

Issue #202, opened 2021-06-15
[Feat]: SSR support


bartlomiejzuber:
> [!NOTE]  
> **Edit:** This feature is under development:
>
> - #1095
>
> _Edited by @davidjerleke 2024-12-26._

---

### Bug is related to

- [x] The Embla Carousel core (all versions)

### Embla Carousel version

- v4.5.3

### Describe the bug

- When carousel is being rendered via SSR there's noticeable flickering effect on first slide on each page refresh (regardless of cache).

### CodeSandbox

- https://codesandbox.io/s/suspicious-burnell-mfs1q

### Steps to reproduce

1. Go to https://mfs1q.sse.codesandbox.io/
2. Refresh the page and look at carousel.

### Expected behavior

- It shouldn't flicker when SSR is used.

### Additional context

![flicker](https://user-images.githubusercontent.com/9931428/122062513-c58a8700-cdef-11eb-995c-6e3d4de82726.gif)

davidcetinkaya:
Hi @bartlomiejzuber,

Thank you for your bug report. This might come as a surprise but this is expected. Embla Carousel is a client side script and won't do anything on the server. This is because Embla needs to read the container and slide dimensions in order to align the carousel correctly, and place some slides in loop positions. This has to be done on the client because element dimensions doesn't exist on the server. So this is what's happening:

- You serve your HTML and CSS from the server.
- The client script mounts and Embla picks up the container and slide dimensions (slide sizes, spacing between slides etc).
- Embla aligns the slides and positions slides needed to create the loop effect.
- Hence, you'll see the flicker when this happens.

One way to solve this would be to hide the carousel with `opacity: 0;` until Embla has initialized. Alternatively, you can add `opacity: 0;` to all the slides if you still want to display the carousel arrows and dots until Embla has initialized. There are probably other ways to do it but I think you get the idea.

I hope this helps.

Best,
David

manojVivek:
@davidcetinkaya Do you think the approach suggested in this issue, helps fix this problem? https://github.com/FormidableLabs/nuka-carousel/issues/786#issue-928155497 ?

davidcetinkaya:
Hi @manojVivek,

Currently, the HTML gets server side rendered and you don't loose any important SEO. I think the flicker thing is solvable, but if this is going to be implemented we need to ask ourselves at what cost? How much will it affect bundle size and what would be a good solution for this?

mariaaraujo:
[...]

PR #1095, opened 2024-12-25
[Feat]: SSR support

davidjerleke:
- Closes: #202.


packages/embla-carousel/src/components/Axis.ts
```diff
@@ -49,6 +50,7 @@ export function Axis(
     cross,
     startEdge,
     endEdge,
+    isRightToLeft,
```
davidjerleke, line 53: May not be necessary anymore. Remove if that’s the case.

davidjerleke: @sarussss thank you for your question. Please move your question to the issue instead and I will answer you there:

- #202

More devs will find the issue rather than this pull request, and this thread is for pull request related discussions only.

Thank you for understanding!

Issue #899, opened 2024-06-07
[Bug]: Offset by 1 issue for canScrollNext when tab size has fractional width


DBulinski:
### Which variants of Embla Carousel are you using?

- [X] embla-carousel **(Core)**
- [X] embla-carousel-react
- [ ] embla-carousel-vue
- [ ] embla-carousel-svelte
- [ ] embla-carousel-autoplay
- [ ] embla-carousel-auto-scroll
- [ ] embla-carousel-solid
- [ ] embla-carousel-auto-height
- [ ] embla-carousel-class-names
- [ ] embla-carousel-fade
- [ ] embla-carousel-docs **(Documentation)**
- [ ] embla-carousel-docs **(Generator)**

### Steps to reproduce

The bug occurs when I have multiple slides on the page, and slides have some specific fractional width (I've identified so far .5px at the end when the number of slides is even and .62px for 3 slides. But I think there is more.

1. Open stackblitz
2. Open carousel in new window
3. Resize the window, so the slide width is rounded to 0.5px
4. Watch the indicator below carousel (Can / Cant). It's the return value from api.canScrollNext()

https://github.com/davidjerleke/embla-carousel/assets/84719303/8c37bb1d-b7b2-4079-9d1a-f8c5cad873ce

### Expected Behavior

It should finish on the last scroll, and doesn't have offset by 1 issue.

### Additional Context

I'm almost sure this is an issue, but would be super grateful if there is something wrong in my code and you can point it. The implementation on stackblitz is a simple copy from https://ui.shadcn.com/docs/components/carousel. If you confirm that this is a bug a can try to fix it, but would need some guidance, what can be an issue :) 

### What browsers are you seeing the problem on?

Chrome, Safari, Microsoft Edge

### Version

v8.1.3

### CodeSandbox

https://stackblitz.com/edit/vite-react-tailwind-jhsdqz?file=src%2FApp.jsx

### Before submitting

- [X] I've made research efforts and searched the documentation
- [X] I've searched for existing issues
- [X] I agree to follow this project's Contributing Guidelines for bug reports

sarussss:
Hi @DBulinski 
I think it might have something to do with [this](https://github.com/davidjerleke/embla-carousel/discussions/346).

davidjerleke:
@DBulinski thanks for the clear bug report. And thanks @sarussss.

The problem could be related to the transition to reading element offsets instead of client rects:

[...]

PR #1096, opened 2024-12-28
#899 - Fix canScroll calculation for fractional slide widths

yasuhiro-yamamoto:
The issue was caused by rounding differences between actual slide widths and offsetWidth calculations:

- When using NodeRects for size calculations, offsetWidth returns integer values even when actual slide widths have decimal points
- This integer rounding caused a 1px discrepancy in the scroll possibility check array when slide widths were fractional
- As a result, canScroll returned incorrect boolean values for certain window/slide sizes
- Added a 1px tolerance filter to prevent unexpected behavior from these sub-pixel differences

This fix ensures consistent canScroll behavior regardless of fractional slide widths by filtering out negligible 1px differences that should not affect scroll possibility determination.

Related to issue #899


davidjerleke:
Hi @yasuhiro-yamamoto,

Thank you for your contribution. I'm working on `v9` at the time of writing so it will take time before I have the chance to check this out. Thanks you for your patience.

**One question:** Regarding steps to reproduce the problem, can you reproduce the problem you describe simply by following the steps in this bug report?

- #899

yasuhiro-yamamoto:
@davidjerleke  Thank you for your response.

Yes, I can reproduce the issue with the steps provided in the Issue. Here are the key points that I've confirmed also 
reproduce the issue in playgrounds/embla-carousel-playground-react:

- When the view displays an even number of slides, the issue occurs if the total number of slides is even (e.g., with flex-basis: 50%, the issue appears with 4 or 6 total slides)
- When the view displays an odd number of slides, the issue occurs if the total number of slides is odd (e.g., with flex-basis: 33.3%, the issue appears with 5 or 7 total slides)

I noticed that my initially proposed code wasn't comprehensive enough and could still cause bugs in some cases, so I've added a commit with additional fixes.

I appreciate your time in reviewing this when possible.

yasuhiro-yamamoto:
Thank you for your feedback! I'm really glad I could help. This is such a great library, looking forward to contributing more!

davidjerleke:
@yasuhiro-yamamoto a bug fix for this was just released in [8.5.2](https://github.com/davidjerleke/embla-carousel/releases/tag/v8.5.2). Thank you a lot for your contribution 🎉!

PR #1230, opened 2025-08-22
[Feat]: Allow AutoScroll speed changes without resetting carousel

Ammar1999y:
It would be nice to have a way to control the carousel animation speed.  
I added this feature in my PR, and it helped in cases like slowing down the transition when the user interacts or when the mouse enters.

- Closes #1229.

### DIFF
--- a/packages/embla-carousel-auto-scroll/src/components/AutoScroll.ts
+++ b/packages/embla-carousel-auto-scroll/src/components/AutoScroll.ts
@@ -25,6 +25,7 @@ export type AutoScrollType = CreatePluginType<
     stop: () => void
     reset: () => void
     isPlaying: () => boolean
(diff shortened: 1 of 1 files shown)


Ammar1999y:
Thanks for the quick response, @davidjerleke 🙏
The suggestions you added were really helpful.
I’ve applied the changes and pushed the update to the branch. 🚀

romanryu:
Is this actually released? I cannot find any mentions of `setSpeed` neither in source files in my node_modules, nor in [npm](https://www.npmjs.com/package/embla-carousel-auto-scroll?activeTab=code), both version `8.6.0`. Could it be a publishing bug?
this is in npm:
image
this is in this repo:
image

davidjerleke:
@romanryu thanks for your question. I decided to release it with the upcoming `v9`. So no, this has not been released yet.

PR #1236, opened 2025-09-04
docs: clarify slidesToScroll=auto behaviour

dsd:
I spent some time trying different approaches to make the carousel dynamically scroll according to how many elements are currently visible in the viewport.

Then I discovered that this option already exists! But the documentation was not making that as clear as I would have liked.

Here's my attempt at improving the documentation.


davidjerleke:
Hi @dsd,
Thanks for your pull request — the clarification is definitely helpful.

**One thought:** should we also clarify that when `slidesToScroll: "auto"` is used, the carousel groups however many slides fit *into the viewport width*, based on their individual sizes?

For example, if slide widths vary — like `50%, 50%, 25%, 25%, 25%, 25%, 33.33%, 33.33%, 33.33%, 80%, 40%` — the carousel would scroll in groups like:

* 2 slides (`50% + 50%`)
* 4 slides (`25% * 4`)
* 3 slides (`33.33% * 3`)
* 1 slide (`80%` — since `80% + 40% > 100%`)
* 1 slide (`40%` — only one left)

So not all groups will contain the same number of slides. Might be worth pointing that out to avoid confusion, especially for those using variable-width slides.

What do you think?

dsd:
OK, now I see why the word "group" was used in the current docs.
How about this revised attempt?

davidjerleke:
@dsd looks good, thank you for your contribution 🎉.

PR #1242, opened 2025-09-10
feat: Configurable settle threshold

matvp91:
This PR makes the threshold used to mark the animation as "settled" configurable.

We initially want to influence the threshold because we'd like to mark the first item in a row as focus after settling. We noticed it takes a while for the event to be fired, as the threshold is rather low.

We kept the current value (0.001) as default.

### DIFF
--- a/packages/embla-carousel-docs/src/content/pages/api/options.mdx
+++ b/packages/embla-carousel-docs/src/content/pages/api/options.mdx
@@ -546,6 +546,15 @@ Enables **infinite looping**. Embla will apply `translateX` or `translateY` to t
 
 ---
 
+### settledThreshold
+
+Type: <BrandPrimaryText>`number`</BrandPrimaryText>
+Default: <BrandSecondaryText>`0.001`</BrandSecondaryText>
+
+Allow the carousel to stop animating when below the threshold.
+
+---
+
 ### skipSnaps
 
 Type: <BrandPrimaryText>`boolean`</BrandPrimaryText>  
--- a/packages/embla-carousel/src/components/Engine.ts
+++ b/packages/embla-carousel/src/components/Engine.ts
@@ -177,7 +177,6 @@ export function Engine(
       previousLocation,
       scrollLooper,
       slideLooper,
-      dragHandler,
       animation,
       eventHandler,
       scrollBounds,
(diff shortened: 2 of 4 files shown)


davidjerleke:
Hi @matvp91 and @thomasdbock,
Thanks a lot for your contribution!

A quick reminder: Please make sure to review the [contributing guidelines](https://github.com/davidjerleke/embla-carousel/blob/master/CONTRIBUTING.md#new-features) before submitting PRs — especially when introducing new features.

In this case, the most relevant section is:

> **`Discuss:`** Create a [new discussion](https://github.com/davidjerleke/embla-carousel/discussions/new?category=ideas) before starting significant work. This helps us align new features with Embla Carousel’s vision.

### Available solution
In case you've missed it, there is an available solution for this here:

- https://github.com/davidjerleke/embla-carousel/discussions/738#discussioncomment-8404744

### Regarding this PR

I understand your goal — to make the threshold for marking an animation as "settled" configurable. You've kept the current default value (`0.001`), which is great for backward compatibility. I think a discussion for this already exists:

- #738

However, there are a couple of concerns:

1. **Bundle size considerations:**
I'm still evaluating what features will make it into v9, and I'm being quite strict about bundle size. Any new option or logic has to justify its weight and complexity, especially if it adds configuration surface that might not be widely used.

2. **Premature scroll completion:**
   Lowering the threshold will likely cause the scroll animation to complete before reaching its actual target. You can see how this is handled here:
  https://github.com/davidjerleke/embla-carousel/blob/e3dbdb33f7b1da756ba2af561c4247c33f816d1a/packages/embla-carousel/src/components/Engine.ts#L192

3. **v9 changes & merge conflicts:**
   The upcoming `v9` introduces significant changes, and merging this into `main` now could cause conflicts or require adjustments later — all of which will be veery time consuming. If you'd like to proceed, please rebase your PR onto the `feature/v9` branch instead, but please be aware of point `1. Bundle size considerations` above.

Thanks and I appreciate your interest in improving Embla Carousel.

matvp91:
Hey @davidjerleke,

Thank you for your detailed response, highly appreciated.

Your suggested solution shall work for us. I'll close this PR.

As you're open to feedback, it might be taken into consideration in v9 that users  don't expect the `settle` event to be delayed (not purposely though). There's probably good reasons I'm not aware of from a technical point of view, but from a user perspective, it feels like it should be there the moment the end user no longer experiences the animation (the delta is too small to be noticed nearing the end).

Happy to give v9 a spin once it's somewhat stable.

All the best!

davidjerleke:
@matvp91 thank you for understanding. I’ve added this to the roadmap so I will consider adding this:

- #841 

Cheers!

PR #1293, opened 2026-02-28
[Bug]: Types - Add `undefined` for `EmblaPluginsType` members

tetarchus:
Adds `undefined` to plugins returned by the `api.plugins()` function.

This ensures that the types are more accurate as a typescript will currently list a plugin as a _definite_ value returned by the `plugins()` function if its module is imported (due to interface merging), however if it is _not_ passed to the embla carousel, it will fail at runtime due to that plugin not being loaded.

The [documentation](https://www.embla-carousel.com/plugins/autoplay/#installation) for calling functions on the plugin uses optional chaining to get around this, however lint rules such as [`@typescript-eslint/no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/) will flag this as not required, and typescript will not alert that this _is_ required - this change will resolve both issues and bring the types in-line with suggested usage to allow editor hints.

### DIFF
--- a/packages/embla-carousel/src/components/Plugins.ts
+++ b/packages/embla-carousel/src/components/Plugins.ts
@@ -23,4 +23,6 @@ export interface EmblaPluginsType {
   [key: string]: CreatePluginType<LoosePluginType, {}> | undefined
 }
 
-export type EmblaPluginType = EmblaPluginsType[keyof EmblaPluginsType]
+export type EmblaPluginType = NonNullable<
+  EmblaPluginsType[keyof EmblaPluginsType]
+>


davidjerleke:
Hi @tetarchus,
Thanks for the contribution and for the detailed explanation of the change — I appreciate the effort to improve the type safety here.

Could you please rebase the PR on the latest `master` branch? Once that’s done, let me know and I’ll have a look as soon as possible.

tetarchus:
@davidjerleke I've updated to the latest `master` branch.

PR #1327, opened 2026-04-02
[Bug]: Eliminate forced reflow during carousel initialization

dlabadini:
## Problem

`NodeHandler.getRects()` unconditionally writes `container.style.transform = 'none'` before reading `offsetWidth`/`offsetHeight`, then restores it. This write-then-read pattern forces a synchronous browser reflow on every `activate()` call — even on initial mount when no transform has been applied yet (`transform` is `''`).

Additionally, `activate()` calls `engine.translate.to()` (which writes `style.transform`) before `engine.resizeHandler.init()` (which reads `offsetWidth`/`offsetHeight`), causing a second forced reflow per carousel init.

On pages with multiple carousels, these compound: each carousel init forces 2+ synchronous reflows, blocking the main thread during hydration.

## Fix

**`getRects`**: Guard the transform write — only mutate `style.transform` when it's actually set to something other than `'none'` or `''`. On initial `activate()`, the transform is empty, so the guard skips the write entirely and the `offsetWidth`/`offsetHeight` reads don't force reflow.

**`activate`**: Move `resizeHandler.init()` and `container.offsetParent` (reads) before `translate.to()` and `scrollOptimizer.optimize()` (writes). Neither read depends on the translate position. This batches all layout reads before any layout writes.

## Impact

- Zero behavioral change — same dimensions, same snap points, same drag bounds
- Eliminates forced reflow on initial carousel mount (the hot path)
- On `reActivate()` after scrolling, the transform guard correctly fires when a transform is present

## Verification

Chrome DevTools Performance panel before/after on a page with 9 Embla carousels:
- Before: "Forced reflow while executing JavaScript" per carousel init, ~354ms Style & Layout
- After: Clean — zero forced reflows from Embla, ~150ms Style & Layout

### DIFF
--- a/packages/embla-carousel/src/components/EmblaCarousel.ts
+++ b/packages/embla-carousel/src/components/EmblaCarousel.ts
@@ -134,18 +134,20 @@ function EmblaCarousel(
     if (!options.active) return
 
     if (!isSsr && ownerWindow) {
+      engine.resizeHandler.init(ownerWindow)
+      const containerVisible = container.offsetParent
+
       engine.translate.to(engine.location)
       engine.scrollOptimizer.optimize(true)
       if (engine.options.loop) engine.slideLooper.loop()
 
       engine.animation.init(ownerWindow)
-      engine.resizeHandler.init(ownerWindow)
       engine.slidesInView.init(ownerWindow)
       engine.slidesHandler.init(ownerWindow)
       engine.slideFocus.init(ownerWindow)
       engine.eventHandler.init(self)
 
-      if (container.offsetParent && slides.length) {
+      if (containerVisible && slides.length) {
         engine.dragHandler.init(ownerWindow)
       }
     }
--- a/packages/embla-carousel/src/components/NodeHandler.ts
+++ b/packages/embla-carousel/src/components/NodeHandler.ts
@@ -64,12 +64,13 @@ export function NodeHandler(root: HTMLElement): NodeHandlerType {
 
     const containerStyle = root ? container.style : { transform: '' }
     const previousTransform = containerStyle.transform
-    containerStyle.transform = 'none'
+    const needsReset = previousTransform && previousTransform !== 'none'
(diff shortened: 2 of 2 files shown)


davidjerleke:
@dlabadini Thanks a lot for taking the time to investigate this and for putting together such a detailed write-up. I appreciate the effort.

That said, I don't think we can make this change.

The key issue is that `transform` can still be applied to the container and slides via CSS rules, even when `element.style.transform` is `''` or `'none'`. Checking only the inline style doesn't tell us whether a transform is actually affecting layout. Because of that, Embla has to set `transform: none` unconditionally just before measuring dimensions and positioning the carousel to guarantee correct measurements. The alternative would be to read the computed style (for example via `getComputedStyle`) to determine whether a transform is applied, but that would itself force a style/layout calculation, so it doesn't avoid the underlying cost.

This is especially relevant with the new [SSR plugin](https://www.embla-carousel.com/docs/plugins/ssr). It applies transforms through CSS rules to eliminate CLS during hydration. In that scenario, skipping the inline `transform = 'none'` write would cause Embla to measure the transformed elements, resulting in incorrect positioning.

So while the proposed optimization looks like it removes forced reflows, it unfortunately changes the behavior in cases where transforms come from CSS rather than inline styles. The reflow is effectively unavoidable if we want to guarantee correct measurements across all supported scenarios.

The best optimization here, and in my opinion the approach that should generally be preferred by devs, is to avoid initializing carousels that are not in view yet. You can still render all the carousel markup, so there is no SEO loss for crawlers, and you can use the SSR plugin to preposition the carousel to avoid layout shifts. Then the actual Embla initialization work only happens when the carousel is needed.

This is one of those edge cases that's easy to miss (and something AI-generated optimizations often overlook): the change appears to improve performance, but it subtly breaks correctness for valid scenarios. In this case, correctness has to take precedence.

Thanks again for digging into this and for the thoughtful proposal.

PR #1328, opened 2026-04-02
[Feat]: Add asyncInit option for non-blocking dimension measurement

dlabadini:
## Motivation

Embla reads `offsetWidth`/`offsetHeight` synchronously during `activate()` to compute slide dimensions. In frameworks like React, these reads happen inside `useEffect` — after React's DOM commit has written class/style changes from other components. This cross-component write-then-read forces a synchronous browser reflow, blocking the main thread.

This is increasingly problematic as pages render more carousels (e.g., discovery pages with 8+ carousels), and as frameworks move toward concurrent/deferred rendering where effects interleave.

Related: #1321 (sync reInit inside ResizeObserver), #1326 (SSR plugin extraction)

## Solution

Add an `asyncInit` option (default: `false`) that uses `ResizeObserver` for initial dimension measurement instead of synchronous `offsetWidth`/`offsetHeight` reads.

When `asyncInit: true`:
1. `activate()` sets up a `ResizeObserver` on the container + slides
2. The observer fires with `borderBoxSize` dimensions (no layout read)
3. Engine is created with those dimensions
4. Carousel becomes interactive

The one-frame delay (~16ms) between mount and interactive is invisible to users — the carousel is already visually rendered via CSS, just not yet draggable/snappable.

When `asyncInit: false` (default): current sync behavior, fully backward compatible.

## Design decisions

- **Opt-in, not default**: Preserves backward compatibility. Consumers that depend on sync API availability after mount are unaffected.
- **ResizeObserver over rAF**: ResizeObserver reports dimensions without triggering layout recalculation. `requestAnimationFrame` would still require `offsetWidth` reads.
- **`borderBoxSize`**: Available in all modern browsers (Chrome 84+, Firefox 92+, Safari 15.4+). More accurate than `offsetWidth` for elements with box-sizing variations.
- **Cleanup via mediaHandlers**: The observer disconnects on `deActivate()` using the existing cleanup infrastructure.

## Test plan

- [ ] Existing tests pass (sync path unchanged)
- [ ] `asyncInit: false` behaves identically to current behavior
- [ ] `asyncInit: true` produces same snap positions after observer fires
- [ ] Carousel is interactive after one frame with `asyncInit: true`
- [ ] `destroy()` during async init cancels observer cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/embla-carousel/src/__tests__/asyncInit.test.ts
+++ b/packages/embla-carousel/src/__tests__/asyncInit.test.ts
@@ -1,0 +1,25 @@
+import EmblaCarousel from '../components/EmblaCarousel'
+import { mockTestElements } from './mocks'
+import { FIXTURE_CONSTRUCTOR_1 } from './fixtures/constructor.fixture'
+
+describe('➡️  asyncInit option', () => {
+  test('Creates engine with sync init by default', () => {
+    const emblaApi = EmblaCarousel(
+      mockTestElements(FIXTURE_CONSTRUCTOR_1),
+      { containScroll: false }
+    )
+    const engine = emblaApi.internalEngine()
+    expect(engine.scrollSnaps.length).toBeGreaterThan(0)
+    emblaApi.destroy()
+  })
+
+  test('Accepts asyncInit option without error', () => {
+    const emblaApi = EmblaCarousel(
+      mockTestElements(FIXTURE_CONSTRUCTOR_1),
+      { containScroll: false, asyncInit: false }
+    )
+    const engine = emblaApi.internalEngine()
+    expect(engine.scrollSnaps.length).toBeGreaterThan(0)
+    emblaApi.destroy()
+  })
+})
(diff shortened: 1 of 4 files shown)


kaelansmith:
Just FYI, I incorporated this change into one of my sites which had a Forced Reflow problem according to Lighthouse, and this seems to have mostly fixed it. Lighthouse still _sometimes_ flags Forced Reflow as an issue depending on the run (previously it was always flagged), but when it does flag it, the amount of Reflow in milliseconds is much lower than before (not sure if that occasional reflow is caused by Embla or something else). I first tried the branch from your other PR https://github.com/davidjerleke/embla-carousel/pull/1327 but that one didn't seem to make a difference with my reflow issue according to Lighthouse.

davidjerleke:
Please see my response here:

- https://github.com/davidjerleke/embla-carousel/pull/1327#issuecomment-5034067526

PR #1351, opened 2026-05-21
[Doc] Update navigation functions to use scroll methods

LucaArgentieri:
Fixed wrong method used goToPrev -> scrollPrev and goToNext -> scrollNext


davidjerleke:
@LucaArgentieri thanks for considering contributions to Embla. `goToPrev` and `goToNext` is the correct methods for the latest `v9` though. In v8 they were called `scrollPrev` and `scrollNext`.

PR #1383, opened 2026-08-26
[Bug]: reInit() is called synchronously inside ResizeObserver callback

raphyabak:
## Summary

- Fixes #1321.

`ResizeHandler`'s `onResize` calls `event.api.reInit()` synchronously from inside the `ResizeObserver` callback whenever a watched node's size changes by 0.5px or more:

```ts
if (diffSize >= 0.5) {
  event.api.reInit()
  break
}
```

`reInit()` can itself change layout (slide translation, etc.), which the observer picks up while still delivering the current notification batch — some browsers report this pattern as a `ResizeObserver loop` warning. This reproduces deterministically in an isolated, framework-free repro (see the issue) by tracking whether `reInit()` runs while the `ResizeObserver` callback is still active.

## Fix

Defer `reInit()` to a `requestAnimationFrame` instead of calling it synchronously, matching the same approach already used for the initial `observe()` calls in this file. To address the point raised in the issue thread about not just trading the warning for avoidable work, this also:

- **Coalesces** multiple resize batches: if further batches arrive before the pending frame fires, they don't schedule additional `reInit()` calls — a burst of `ResizeObserver` notifications settles into a single reinitialize.
- **Cancels** the pending frame on `destroy()`, so a resize that fires just before teardown can't call `reInit()` against an already-destroyed instance.

## Test plan

- Added two tests to `resize-ltr.test.ts`:
  - Three resize batches in a row schedule only one deferred `reInit()`, and it doesn't run until the frame fires.
  - A pending reinitialize is cancelled (via `cancelAnimationFrame`) when the carousel is destroyed first, and `reInit()` is not called even if the frame still fires afterward.
- Verified the regression: reverting the source change turns both new tests from passing into failing — `reInit()` runs synchronously instead of being deferred, and `cancelAnimationFrame` is never called on destroy.
- Full package suite passes: `yarn jest --config jest.config.js` — 572 passing, all 3 resize test files (ltr/rtl/vertical, 23 tests) included, no changes needed to the existing tests.
- `eslint`, `tsc --noEmit`, and `prettier --check` all pass on the changed files.

### DIFF
--- a/packages/embla-carousel/src/__tests__/resize-ltr.test.ts
+++ b/packages/embla-carousel/src/__tests__/resize-ltr.test.ts
@@ -155,4 +155,83 @@ describe('➡️  Resize - Horizontal LTR', () => {
       expect(callback).toHaveBeenCalledTimes(0)
     })
   })
+
+  describe('When multiple resize batches arrive before the deferred reinitialize runs', () => {
+    // window.requestAnimationFrame is already a jest.fn() (see
+    // mocks/requestAnimationFrame.mock.ts) that invokes its callback
+    // synchronously. mockImplementationOnce overrides it for exactly the one
+    // call under test and then automatically falls back to that default
+    // synchronous behavior, so nothing needs to be restored afterwards - and
+    // unrelated requestAnimationFrame calls elsewhere in the engine are
+    // unaffected.
+    function captureNextAnimationFrame(): { frame?: FrameRequestCallback } {
+      const captured: { frame?: FrameRequestCallback } = {}
+      ;(window.requestAnimationFrame as jest.Mock).mockImplementationOnce(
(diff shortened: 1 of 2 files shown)


manualrabo:
Hi @davidjerleke our web app and customers are impacted due to this issue.
Can we please merge this and release?

~ Cheers

davidjerleke:
Hey @manualrabo — I understand the impact, but this is an OSS project, so it’s not exactly paying my bills. I have very limited time to work on it at the moment, so unfortunately you’ll have to wait until I can get to it.

I’ll take a look as soon as I have some time. Thanks for your patience.

manualrabo:
Looking forward to it, appreciate the OSS efforts, Thank you 🙏

Issue #1389, opened 2026-09-08
Accessibility plugin hides visible slides when a snap group is smaller than the viewport fits


tobiCodable:
## Description

`embla-carousel-accessibility` decides slide suppression from snap-group membership rather than actual visibility. When a snap group covers fewer slides than the viewport shows, fully visible slides are given `aria-hidden="true"` and their focusable children `tabindex="-1"`.

`updateSlides` in `components/Accessibility.ts`:

```js
const slidesInSnap = slidesBySnap[snapIndex];
const isActive = snapIndex === emblaApi.selectedSnap();

slidesInSnap.forEach(slideIndex => {
  // …
  slideAttributes.toggle('aria-hidden', !isActive && !hasFocusedElement);
  slideFocusNodes.forEach(slideFocusNode => {
    const tabindex = !isActive ? '-1' : prevTabIndex;
    focusAttributes.toggle('tabindex', tabindex);
  });
});
```

`slidesBySnap` answers "which slides does this snap scroll to", not "which slides can the user see right now". The two agree while the overflow is large. They diverge when the carousel only just overflows its viewport.

## Reproduction

Six slides, five per view (CSS flex-basis, so each slide is a fifth of the viewport minus gap and peek), a `2rem` gap, viewport 1440px, `loop: true`, `align: "center"`, `slidesToScroll` left at its default.

Embla produces two snaps and groups the slides `[[0,1,2],[3,4,5]]`. Measured DOM state at the first snap:

| slide | fully inside viewport | `aria-hidden` | child button `tabindex` | slide `aria-label` |
|---|---|---|---|---|
| 1 | no (centre align) | – | – | Slides 1-3 of 6 (group 1 of 2) |
| 2 | yes | – | – | Slides 1-3 of 6 (group 1 of 2) |
| 3 | yes | – | – | Slides 1-3 of 6 (group 1 of 2) |
| 4 | **yes** | **true** | **-1** | Slides 4-6 of 6 (group 2 of 2) |
| 5 | **yes** | **true** | **-1** | Slides 4-6 of 6 (group 2 of 2) |
| 6 | no (clipped) | true | -1 | Slides 4-6 of 6 (group 2 of 2) |

Slides 4 and 5 are wholly on screen. A sighted mouse user can click their buttons; a keyboard user cannot reach them, and a screen reader does not announce them.

## Impact

[...]

PR #1397, opened 2026-09-19
suggested fixes for issue #1389 (re: Accessibility plugin hides visib…

humonious17:
…le slides when a snap group is smaller than the viewport fits)


humonious17:
@davidjerleke can you please review the code and lmk if there are any changes required against this issue's fix.

